PR #36, opened 2026-04-07
Improve error message when argument order matches but type annotation is incorrect

Darshan808:
### Description
- Added a new error message:
  ```ts
  incorrectType: 'Activation argument "{{ arg }}" has incorrect type annotation "{{ type }}". Expected type compatible with "{{ expected }}".',
  ```
- Added logic to distinguish between order mismatch and incorrect type annotation.
- Removed conditional type-checking logic:
    ```ts
      const hasTypeInfo = actualParamTypes.some(t => t !== null);
      if (hasTypeInfo) {
    ```
- Previously, type validation was only performed when type information was present.
- Updated behavior to enforce that all activation parameters must have valid type annotations. If a parameter is missing a type (or null), an error is now raised.
- This change was motivated by a failing test case that was incorrectly passing:
  ```ts
    activate: (app: JupyterFrontEnd, tracker: null) => {
  ```

### DIFF
--- a/src/rules/plugin-activation-args.ts
+++ b/src/rules/plugin-activation-args.ts
@@ -155,7 +155,7 @@ const jupyterPluginActivationArgs = createRule({
       wrongArgumentCount:
         'Expected {{ expected }} activation arguments (app + {{ tokenCount }} tokens), got {{ actual }}.',
       unresolvableTokenType:
-        'Token "{{ token }}" type could not be resolved — package may be unbuilt. Build it for accurate type checking.'
+        'Token "{{ token }}" type could not be resolved. The package may be unbuilt or type checking may not be configured.'
     },
     fixable: 'code',
     schema: [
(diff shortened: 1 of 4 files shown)


src/rules/plugin-activation-args.ts
```diff
@@ -288,8 +290,12 @@ const jupyterPluginActivationArgs = createRule({
         }
       }
       // Without a checker we cannot resolve namespace patterns like
-      // `IDebugger.ISidebar` ↔ `IDebuggerSidebar`.
-      if (!checker && paramType && paramType.includes('.')) return [true, false];
+      // `IDebugger.ISidebar` ↔ `IDebuggerSidebar`, so we cannot distinguish
+      // a genuine mismatch from a valid aliasing convention. Treat as a match
+      // to avoid false positives; the exact-name check above already handles
+      // the common case. Users wanting full type-aware checks should configure
+      // `parserOptions.project`.
```
Darshan808: After some thought, I think that if type-checking is not configured (even though we recommend it in the docs), we should ignore these mismatches since there is no reliable way to verify them.

What do you think, @krassowski? I’ve also modified docs to include this behavior in the PR.

krassowski: I think we don't want to give a false sense of security. If the plugin is misconfigured then I would err on side of logging a warning saying "hey we need access to types to resolve it properly" than ignoring it. On the other hand, if we can resolve everything without full type configuration then maybe it is fine to not prompt for additional config.  So I am not convinced we should ignore it.

Darshan808: Hmm, makes sense. Yes let's log `'Token "{{ token }}" type could not be resolved. The package may be unbuilt or type checking may not be configured.'`, if we need but cannot perform type-aware checking.

krassowski (approved):
Thanks!

PR #37, opened 2026-04-07
Report empty descriptions

### LINKED ISSUE #28: Make `@jupyter/plugin-description` rule more verbose.
### Description
Originally posted by copilot on: https://github.com/jupyterlab/eslint-plugin/pull/27#discussion_r2936234968
> hasDescriptionProperty() returns false for an empty-string literal, but the rule always reports messageId: 'missingDescription'. That makes the error message inaccurate and leaves the emptyDescription message unused. Consider distinguishing between “missing” vs “present but empty” (e.g., return a tri-state / enum from the helper) and report emptyDescription when the property exists but is an empty/whitespace-only literal.

Darshan808:
### Description
- Updated `jupyter/plugin-description` rule to also report empty descriptions, in addition to missing descriptions.

### DIFF
--- a/src/rules/plugin-description.ts
+++ b/src/rules/plugin-description.ts
@@ -7,10 +7,19 @@ import { createRule } from '../utils/create-rule';
 import { TSESTree } from '@typescript-eslint/types';
 import { getJupyterPluginKind, getPluginId } from '../utils/plugin-utils';
 
+type DescriptionStatus = 'missing' | 'empty' | 'present';
+
 /**
- * Checks if an object expression has a description property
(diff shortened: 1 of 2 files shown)


krassowski (commented):
Looks good to me, will double check with copilot

krassowski (approved):
Feel free to address od ignore the suggestions, no preference from my side.

Darshan808:
I'm confused regarding the failure in downstream tests on JupyterLab, even though the lint tests pass on the `main` branch of core.

PR #38, opened 2026-04-08
Fix downstream lint config

Darshan808:
Identified in #37 

### Description

Downstream linting wasn't running completely because `parserOptions.project` was not set in our downstream linting config.  
As a result, there was no type-checking, and type-aware linting rules were not running properly.

This PR fixes the issue by supplying `parserOptions.project` in the linting configs of downstream projects, enabling proper type-aware linting.


krassowski (approved):
Thanks!

PR #40, opened 2026-04-12
Add `jupyter/token-format` to enforce plugin token naming convention

### LINKED ISSUE #39: Add rule to enforce plugin token naming convention
> Several packages are currently using a non-standard format for plugin token symbols instead of the expected `package:TokenSymbol`.  (token part to be a valid JS identifier) convention. This is causing the cursor ( in plugin playground insertion ) to remain disabled for these plugins.

 _Originally posted by @MUFFANUJ  in [jupyterlite/ai/issues/311](https://github.com/jupyterlite/ai/issues/311)_

### Description
Plugin token IDs follow a common convention but are not currently validated. A downstream tool [plugin-playground](https://github.com/jupyterlab/plugin-playground) assumes the token name is a valid JavaScript identifier, which can lead to issues if not followed.

Add a rule `jupyter/token-format` to:
- Ensure token matches `<package>:<valid JS identifier>`

Darshan808:
### Description

Added a new rule `jupyter/token-format` that ensures JupyterLab Token ids follow the `<package>:<TokenSymbol>` naming convention where TokenSymbol (part after the first ':') must be a valid JavaScript identifier.

### DIFF
--- a/eslint.downstream.config.mjs
+++ b/eslint.downstream.config.mjs
@@ -29,7 +29,8 @@ export default [
     rules: {
       'jupyter/command-described-by': 'error',
       'jupyter/plugin-activation-args': 'error',
-      'jupyter/plugin-description': 'error'
+      'jupyter/plugin-description': 'error',
+      'jupyter/token-format': 'error'
     },
     languageOptions: {
       parser: resolvedParser,
@@ -55,7 +56,8 @@ export default [
     rules: {
       'jupyter/command-described-by': 'error',
       'jupyter/plugin-activation-args': 'error',
-      'jupyter/plugin-description': 'error'
+      'jupyter/plugin-description': 'error',
+      'jupyter/token-format': 'error'
     },
     languageOptions: {
       parser: resolvedParser,
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,7 +14,7 @@ const plugin = {
     'plugin-activation-args': pluginActivationArgs,
     'command-described-by': commandDescribedBy,
     'plugin-description': pluginDescription,
-    'token-format': tokenFormat
+    'token-format': tokenFormat,
     'no-untranslated-string': noUntranslatedString
   },
   configs: {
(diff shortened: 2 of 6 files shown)


Darshan808:
Jupyterlab contains one token with `#` in between:
```ts
  readonly stop = new Token<void>('@jupyterlab/application:Router#stop');
```
The downstream tests should fail. Should this be allowed ? Should we change it in jupyterlab or should it just be ignored?

krassowski:
>   readonly stop = new Token<void>('@jupyterlab/application:Router#stop');

[This](https://github.com/jupyterlab/jupyterlab/blob/2e07b66a58c89cdaab9d9b98ba618f7525151c1f/packages/application/src/router.ts#L57-L61) looks like a different use of tokens. This token is never used in a plugin. I think this is fine in that specific case.

> Should this be allowed ? Should we change it in jupyterlab or should it just be ignored?

I think we could allow it or add an ignore statement in JupyterLab. Although we don't do it in core, it might make sense to have a token in a namespace say:

```ts
export namespace MyNamespace {
  export const myToken = Token<someType>("MyNamespace.myToken");
}
```

It is less clear if we should be using `MyNamespace.myToken` or `MyNamespace#myToken`.

I know that it's been 8 years and it is a long shot, but @afshin do you remember the reasoning behind the choice of `#` in https://github.com/jupyterlab/jupyterlab/pull/3687?

krassowski (approved):
I would say let's ship this and gather feedback from users. It looks like 99.9% use case does not involve `#` or `.` in token names. We can add support for it later.

afshin:
> I know that it's been 8 years and it is a long shot, but @afshin do you remember the reasoning behind the choice of # in jupyterlab/jupyterlab#3687?

Hi! I looked at this PR and I think I wanted to draw a distinction that this is just a plain token, not for use in the plugin infrastructure. Additionally it is exposed off of a class instance and it is meant to give a clue about where or what may have gone wrong. We don't really pass tokens around as values in extensions so this was different.

I don't have a strong feeling about keeping this outlier.

krassowski:
Yes, I agree that a token that is not used in a plugin could be ignored altogether. I am not sure how involved would be inferring that from static analysis (especially given that this is often cross-file); it might _not_ be worth implementing an exception if that was to lead to false negatives.

afshin:
Now that symbol is truly ubiquitous we might even discard this token in favor of a symbol and just side-step the problem of how to lint it.

PR #41, opened 2026-04-12
Add a new rule `jupyter/no-untranslated-string`

Darshan808:
### Description
Add a new rule `jupyter/no-untranslated-string` which require user-facing string literals to be wrapped in a translation call.

The rule reports raw string literals (and template literals without expressions) in the following positions:

#### 1. `commands.addCommand()` properties

The `label`, `caption`, and `description` properties of a command registration must not contain bare string literals.

#### 2. `element.setAttribute()` with accessibility attributes

The second argument of `setAttribute` must not be a raw string when the attribute is `aria-label`, `aria-description`, or `title`.

#### 3. Direct assignment to `title` and `ariaLabel`

Assigning a raw string literal to `element.title` or `element.ariaLabel` is flagged.

Feedbacks on the approach is welcome.

### DIFF
--- a/src/rules/no-untranslated-string.ts
+++ b/src/rules/no-untranslated-string.ts
@@ -1,0 +1,156 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+import { TSESTree } from '@typescript-eslint/types';
+import { isAddCommandCall } from '../utils/commands';
+import { getObjectProperties } from '../utils/plugin-utils';
+import { createRule } from '../utils/create-rule';
+
+/**
+ * Returns true if the node is a non-empty raw string literal that should be
+ * wrapped in a translation call. Handles:
+ *   - String Literal: 'string' or "string"
+ *   - TemplateLiteral with no expressions: `string`
+ *   - Concise ArrowFunctionExpression whose body is one of the above
+ */
+function isRawStringNode(node: TSESTree.Node): boolean {
+  if (node.type === 'Literal') {
+    return typeof node.value === 'string' && node.value.length > 0;
+  }
+  if (node.type === 'TemplateLiteral') {
+    if (node.expressions.length > 0) {
+      return false;
+    }
+    const cooked = node.quasis.map(q => q.value.cooked ?? '').join('');
+    return cooked.length > 0;
+  }
+  if (
+    node.type === 'ArrowFunctionExpression' &&
+    node.body.type !== 'BlockStatement'
+  ) {
+    return isRawStringNode(node.body);
+  }
+  return false;
+}
+
+function isSetAttributeCall(node: TSESTree.CallExpression): boolean {
+  if (node.callee.type === 'MemberExpression') {
+    const callee = node.callee;
+    if (
+      callee.property.type === 'Identifier' &&
+      callee.property.name === 'setAttribute'
+    ) {
+      return true;
+    }
+  }
+  return false;
+}
+
+const MONITORED_COMMAND_PROPS = ['label', 'caption', 'description'];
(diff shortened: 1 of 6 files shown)


krassowski:
That's great! These are all true-positive matches that we need to fix:

```
/home/runner/work/eslint-plugin/eslint-plugin/jupyterlab/packages/debugger/src/model.ts
Error:   168:18  error  Assignment to "title" has an untranslated string literal. Wrap it with trans.__()  jupyter/no-untranslated-string

/home/runner/work/eslint-plugin/eslint-plugin/jupyterlab/packages/filebrowser-extension/src/index.ts
Error:   1360:12  error  Command property "label" has an untranslated string literal. Wrap it with trans.__()  jupyter/no-untranslated-string

/home/runner/work/eslint-plugin/eslint-plugin/jupyterlab/packages/notebook/src/widget.ts
Error:   1661:41  error  setAttribute("aria-label", ...) has an untranslated string literal. Wrap the value with trans.__()  jupyter/no-untranslated-string
```

```
/home/runner/work/eslint-plugin/eslint-plugin/notebook/packages/application/src/panelhandler.ts
Error:   93:31  error  Assignment to "title" has an untranslated string literal. Wrap it with trans.__()  jupyter/no-untranslated-string
```

krassowski:
Three broad categories that we could also check:

1. For lumino Widgets:

- `Widget.title.label` and `Widget.title.caption`:
  - https://lumino.readthedocs.io/en/latest/api/classes/widgets.Widget-1.html#title
  -  https://github.com/jupyterlab/jupyterlab/blob/2e07b66a58c89cdaab9d9b98ba618f7525151c1f/packages/debugger/src/panels/sources/index.tsx#L33

2. JupyterLab UI helper dialogs and notifications, e.g.:

```
showDialog({
  title: "Confirm"
});
```

3. Strings in JSX/TSX syntax

---

Additionally, we could forbid concatenation inside translation wrappers, but this could be a separate enhancement in a new rule:

```
trans.__("Delete " + fileName) ❌
trans.__("Delete %1", fileName) ✅
```

src/rules/no-untranslated-string.ts
```diff
@@ -0,0 +1,156 @@
...
+      callee.property.type === 'Identifier' &&
+      callee.property.name === 'setAttribute'
+    ) {
+      return true;
+    }
+  }
+  return false;
+}
+
+const MONITORED_COMMAND_PROPS = ['label', 'caption', 'description'];
```
krassowski: I think commands do not have `description`, they have `usage`

```suggestion
const MONITORED_COMMAND_PROPS = ['label', 'caption', 'usage'];
```

https://lumino.readthedocs.io/en/latest/api/interfaces/commands.CommandRegistry.ICommandOptions.html#usage

Darshan808:
> For lumino Widgets:
> - Widget.title.label and Widget.title.caption:

Seems like we caught 2 more errors from this one.

website/docs/rules/index.md
```diff
@@ -18,6 +19,7 @@ The plugin ships with a recommended configuration that enables all current rules
 | --- | --- |
 | [jupyter/plugin-activation-args](./plugin-activation-args) | `error` |
 | [jupyter/command-described-by](./command-described-by) | `warn` |
+| [jupyter/no-untranslated-string](./no-untranslated-string) | `warn` |
```
krassowski, line 22: Follow-up suggestion: maybe we could make it "error" if `@jupyterlab/translator` is imported as that clearly indicates that extension wants to use translation. For extensions which do not depend on translator maybe we could disable this rule automatically to avoid providing irrelevant suggestions.

krassowski (approved):
Thank you! I think we can merge this and concurrently fix the upstream issues.

PR #44, opened 2026-04-16
Add `jupyter/no-translation-concatenation`

### LINKED ISSUE #43: Forbid concatenation inside translation wrappers
Additionally, we could forbid concatenation inside translation wrappers, but this could be a separate enhancement in a new rule:

```
trans.__("Delete " + fileName) ❌
trans.__("Delete %1", fileName) ✅
```

_Originally posted by @krassowski in https://github.com/jupyterlab/eslint-plugin/issues/41#issuecomment-4236498970_

Darshan808:
### Description
This rule reports any `+` binary expression passed as an argument to a translation method (`__`) called on a recognized translation bundle:

### DIFF
--- a/eslint.downstream.config.mjs
+++ b/eslint.downstream.config.mjs
@@ -29,7 +29,8 @@ export default [
     rules: {
       'jupyter/command-described-by': 'error',
       'jupyter/plugin-activation-args': 'error',
-      'jupyter/plugin-description': 'error'
+      'jupyter/plugin-description': 'error',
+      'jupyter/no-translation-concatenation': 'error'
     },
     languageOptions: {
       parser: resolvedParser,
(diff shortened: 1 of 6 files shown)


krassowski:
Ok so the usage in JupyterLab which attempts to capitalize by splitting and the concatenating string is correctly flagged as incorrect, if anything we should be using locale-aware capitalization methods.

The usage flagged in `packages/application/src/connectionlost.ts` seems to be fine:

```ts
  const networkMsg = trans.__(
    'A connection to the Jupyter server could not be established.\n' +
      'JupyterLab will continue trying to reconnect.\n' +
      'Check your network connection or Jupyter server configuration.\n'
  );
```

At least on Binder it looks that the translation machinery is able to handle concatenation of multiple raw strings if these always produce the same result:

image

Darshan808:
> The usage flagged in packages/application/src/connectionlost.ts seems to be fine:

This pattern isn't flagged now.

krassowski (approved):
Copilot was really annoyed by lack of documentation here to leave as much as three comments on the same topic 😆. I would have let it pass but anyways it looks good now, thanks!

PR #45, opened 2026-04-16
Add jupyterlite to downstream tests.

### LINKED ISSUE #42: Add jupyterlite to downstream testing
### Problem

Next to JupyterLab and Notebook, JupyterLite is the major frontend that include many plugins. We should make sure any new rules or rule changes added here do not adversely impact it.

### Proposed Solution

Now that https://github.com/jupyterlite/jupyterlite/pull/1917 got merged, add jupyterlite to downstream testing.

### Additional context

None

Darshan808:
### Description

Add `jupyterlite` to downstream tests.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -121,7 +121,7 @@ jobs:
       - name: Install and build jupyterlite
         working-directory: jupyterlite
         run: |
-          python -m pip install .
+          python -m pip install --group build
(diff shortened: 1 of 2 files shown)


krassowski (approved):
LGTM! If we end up adding more packages, maybe we could simplify the workflow file to use a single job parametrized by a custom martix, but for n=3 this is fine.

PR #49, opened 2026-04-20
Improve `@jupyter/no-untranslated-string` for `JSX`

Darshan808:
### Description

- Fixed false positives on non-accessibility JSX attribute values (e.g. `className={'...'}`, `id={'...'}`) — these are no longer flagged  
- Still flags monitored accessibility attributes in JSX: `aria-label`, `aria-description`, `title`  
- Skips punctuation/symbol-only JSX text (e.g. `,`, `.`, `' '`) — only flags text containing actual letters  
- Do not flag identifier expressions in JSX text content (e.g. `{myVar}`)

### DIFF
--- a/src/rules/no-untranslated-string.ts
+++ b/src/rules/no-untranslated-string.ts
@@ -299,6 +299,24 @@ const noUntranslatedString = createRule({
         }
       },
 
+      // Accessibility attribute with a plain string: <span aria-label="text" />
+      JSXAttribute(node) {
+        if (!node.value || node.value.type === 'JSXExpressionContainer') {
+          return;
+        }
+        const attrName =
+          node.name.type === 'JSXIdentifier' ? node.name.name : null;
+        if (!attrName || !MONITORED_JSX_ATTRS.includes(attrName)) {
+          return;
+        }
+        if (isRawStringNode(node.value)) {
+          context.report({
+            node: node.value,
+            messageId: 'untranslatedJsxText'
+          });
+        }
+      },
+
       // Raw text between JSX tags: <span>Untranslated text</span>
       JSXText(node) {
         if (node.value.trim().length > 0 && (enforcePunctuation || hasLetters(node.value))) {
@@ -332,7 +350,7 @@ const noUntranslatedString = createRule({
         }
         if (isRawStringNode(node.expression)) {
           const value = getRawStringValue(node.expression);
-          if (value !== null && (enforcePunctuation ? value.length > 0 : hasLetters(value))) {
+          if (value !== null && (enforcePunctuation ? value.trim().length > 0 : hasLetters(value))) {
             context.report({
               node: node.expression,
               messageId: 'untranslatedJsxText'
--- a/src/rules/no-untranslated-string.ts
+++ b/src/rules/no-untranslated-string.ts
@@ -122,11 +122,23 @@ const noUntranslatedString = createRule({
       untranslatedJsxText:
         'JSX text content has an untranslated string literal. Wrap it with {trans.__(...)}'
     },
-    schema: []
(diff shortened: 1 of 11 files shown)


tests/jupyter-no-untranslated-string.test.ts
```diff
@@ -247,7 +247,13 @@ jsxRuleTester.run('no-untranslated-string (JSX)', noUntranslatedString, {
...
     // --- JSX: translated expression ---
     { code: `const el = <span>{trans.__('Error message:')}</span>;` },
-    { code: `const el = (\n  <div>\n    <span>{trans.__('Label')}</span>\n  </div>\n);` }
+    { code: `const el = (\n  <div>\n    <span>{trans.__('Label')}</span>\n  </div>\n);` },
+    { code: `<div className={'normal-class-string'} />` },
+    { code: `<div id={'my-id'} />` },
+    { code: `<span aria-label={trans.__('Close')} />` },
+    // Punctuation-only JSX text should not be flagged
+    { code: `<span>{','}</span>` },
+    { code: `<span>{' + '}</span>` }
```
krassowski, line 256: This is not obvious. There are different symbols for commas in other languages (https://en.wikipedia.org/wiki/Comma#Languages_other_than_English), decimal separator varies widely (https://en.wikipedia.org/wiki/Decimal_separator#/media/File:DecimalSeparator.svg), and even plus sign is written differently in at least on place (https://en.wikipedia.org/wiki/Plus_and_minus_signs#Alternative_plus_sign).

That said in 80% case it is ok to leave punctuation untranslated, and if this helps with adoption of this rule then excluding it makes sense. Maybe it could be an option for the rule configuration so that we can possibly enable it later?

krassowski: Such option could be added in a follow-up PR, if you choose so let's open an issue to track it.

src/rules/no-untranslated-string.ts
```diff
@@ -280,11 +301,38 @@ const noUntranslatedString = createRule({
...
+          }
+          return;
+        }
+        if (node.expression.type === 'Identifier') {
           context.report({
             node: node.expression,
             messageId: 'untranslatedJsxText'
           });
+          return;
+        }
```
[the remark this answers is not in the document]
krassowski: Yes, I'm worried about false positives here too. I am not sure if we could reliably detect: 

```tsx
const text = trans.__('my text')
<div>{text}</div>
```

because it would be false positive even if we only checked for identifiers with string type. We would need to make translation wrappers return a subtype of strings which may be hard to implement (it would be easy in Python as we could flag `LiteralString` https://peps.python.org/pep-0675/ but AFAIK TypeScript does not support that https://github.com/microsoft/TypeScript/issues/51513).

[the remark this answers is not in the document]
krassowski: I guess we could have this as an option, but I would suggest defaulting it to disabled.

tests/jupyter-no-untranslated-string.test.ts
Darshan808: > Maybe it could be an option for the rule configuration so that we can possibly enable it later?

This sounds good. Added an option `enforcePunctuation` which defaults to false.

Darshan808:
Hmm, I don't see a re-request review button for copilot!

krassowski (approved):
Thank you @Darshan808!

krassowski:
Before you merge, you just update the documentation? 

https://github.com/jupyterlab/eslint-plugin/blob/5a745e75d77a75bb65d4e81680a6b7151be5189f/website/docs/rules/no-untranslated-string.md?plain=1#L95-L97

PR #55, opened 2026-05-03
Add `no-schema-enum` rule: forbid enum in settings schema, suggest `oneOf`  

### LINKED ISSUE #52: Rule proposal: forbid use of `enum` in settings schema, suggest `oneOf` instead
### Problem

In JupyterLab/Notebook v7+, settings JSON schema one can use either `enum` of `oneOf`. The former does not allow user-facing labels to differ from value used in code, and does not allow for translations.

### Proposed Solution

Enforce that in JSON schema files used by settings `oneOf` is used instead of `enum`.

### Additional context

This came up numerous times during review:
- https://github.com/jupyterlab/jupyterlab/pull/18237#discussion_r3160980223
- https://github.com/jupyterlab/jupyterlab/pull/17547#discussion_r2077962168

etc

Darshan808:
### Description
- New rule `jupyter/no-schema-enum`, visits `JSONProperty` nodes via `jsonc-eslint-parser`, only activates for files under a `schema/` directory

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,6 +3,7 @@
  * Distributed under the terms of the Modified BSD License.
  */
 
+import * as jsoncParser from 'jsonc-eslint-parser';
 import pluginActivationArgs from './rules/plugin-activation-args';
 import commandDescribedBy from './rules/command-described-by';
 import pluginDescription from './rules/plugin-description';
@@ -35,9 +36,8 @@ const plugin = {
         }
       },
       {
-        // Requires jsonc-eslint-parser configured for JSON files in the
-        // consumer's ESLint config.
         files: ['**/schema/*.json'],
+        languageOptions: { parser: jsoncParser },
         rules: {
           'jupyter/no-schema-enum': 'warn'
         }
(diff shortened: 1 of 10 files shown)


src/index.ts
```diff
@@ -17,7 +18,8 @@ const plugin = {
     'plugin-description': pluginDescription,
     'no-translation-concatenation': noTranslationConcatenation,
     'token-format': tokenFormat,
-    'no-untranslated-string': noUntranslatedString
+    'no-untranslated-string': noUntranslatedString,
+    'no-schema-enum': noSchemaEnum
```
krassowski: should it be added to recommended with `warn` level?

krassowski (approved):
This did highlight a few places where options could have been translated if they had `oneOf`:

Image

In other places the existing translations assumed that they need to mention English code name:

Image

but that can be changed downstream later

src/index.ts
Darshan808: That'll be good. But consumers need to have json parser just like we expect them to have typescript parser.
```js
  {
    files: ['**/schema/*.json'],
    languageOptions: {
      parser: jsoncEslintParser
    },
    rules: {
      'jupyter/no-schema-enum': 'error'
    }
  },
```
Or we could bundle the parser as a direct dependency making `jsonc-eslint-parser` a runtime dependency. What do you say?

krassowski: I think it might be ok to add `jsonc-eslint-parser` as a runtime dependency.

PR #57, opened 2026-05-04
Remove `tsconfig.test.json`

Darshan808:
## Fixes #29


krassowski (approved):
I suppose this is ok, but I am also not sure if we need to change it.

Darshan808:
Using `"commonjs"` with `"moduleResolution": "node"` still works, but TypeScript now flags it as deprecated and the editor keeps showing warnings.
I thought it would be better to upgrade to "node16" now.

PR #60, opened 2026-05-09
Add `jsonc-eslint-parser` as runtime dep

Darshan808:
Discussed in https://github.com/jupyterlab/eslint-plugin/pull/55#discussion_r3180413560 as adding `jsonc-eslint-parser` as a runtime dependency, but it was mistakenly added only as a dev dependency.  
This PR fixes that.
Found out while integrating into downstream projects.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "BSD-3-Clause",
       "dependencies": {
         "@typescript-eslint/utils": "^8.54.0",
-        "jsonc-eslint-parser": "^3.1.0"
+        "jsonc-eslint-parser": "^2.4.0"
       },
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


krassowski:
It looks like CI is failing with:

```
Error: src/index.ts(6,30): error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("jsonc-eslint-parser")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts', or add the field `"type": "module"` to '/home/runner/work/eslint-plugin/eslint-plugin/package.json'.
```

krassowski (approved):
Thanks!

PR #73, opened 2026-07-04
Add `galata-prefer-filebrowser-helper` rule

Darshan808:
## Summary

First in a series of planned `galata-*` rules that encourage using Galata helpers instead of raw Playwright selectors in UI tests.

This rule detects common file browser selectors, including:

- `.jp-DirListing*`
- `.jp-BreadCrumbs*`
- `"File Browser Section"`
- `.ipynb` text matches

and recommends the corresponding Galata helpers instead:

- `page.filebrowser.open()`
- `page.filebrowser.openHomeDirectory()`
- `page.notebook.openByPath()`

## Details

- Detects both `page.click(selector)` and `page.locator(selector).click()` patterns
- Supports template literals by matching only their static portions
- Ignores dynamic selectors that cannot be analyzed reliably
- Provides suggestions only (no autofix)
- Enabled as `warn` in the recommended configuration for test files


Darshan808:
Should we continue with separate, focused rules (`galata-prefer-filebrowser-helper`, `galata-prefer-menu-helper`, etc.), or would it be better to have a single umbrella rule such as `prefer-galata-helpers` that reports all of these patterns?

Darshan808:
It did flag many existing issues. But I see some cases not yet flagged.
```ts
      await page.click('#filebrowser >> .jp-BreadCrumbs-home', { // flagged
        timeout: 5000
      });
      await page.dblclick(`#filebrowser >> text=${tmpPath}`); // Not flagged
```

Darshan808:
Most flagged issues are correct.

A few misleading ones come from cases where we flag clicks on `*.ipynb` assuming the user is opening a file.

Correctly flagged (the user is opening a notebook):

```ts
await page.dblclick('text=Julia.ipynb');
```

Incorrectly flagged (these should not be flagged as `page.notebook.openByPath(path)` because they are not opening a notebook):

```ts
await page.locator('div.lm-TabBar-tabLabel >> text=Notebook.ipynb').click();

await page.click('div[role="main"] >> text=simple_notebook.ipynb', {
  button: 'right'
});
```

MUFFANUJ:
This looks good!, few things - the description says fixes we could change that since we dont want to close the issue it this PR adds one of the rules.

The rule catches this direct raw file-browser open shape:

[`export_notebook.test.ts#L18-L21`](https://github.com/jupyterlab/jupyterlab/blob/6f058ba4d5348b5e4570b44f4639cfc2037ef7f9/galata/test/documentation/export_notebook.test.ts#L18-L21)

```ts
await page.dblclick(
  '[aria-label="File Browser Section"] >> text=notebooks'
);
await page.dblclick('text=Lorenz.ipynb');
```
But it misses the equivalent chained-locator form:
[export_notebook.test.ts#L44-L48](https://github.com/jupyterlab/jupyterlab/blob/6f058ba4d5348b5e4570b44f4639cfc2037ef7f9/galata/test/documentation/export_notebook.test.ts#L44-L48)
```ts
await page
  .locator('[aria-label="File Browser Section"]')
  .getByText('notebooks')
  .dblclick();
await page.getByText('Lorenz.ipynb').dblclick();
```
The second block is still manually opening the notebooks folder through the file browser. The rule reports the `Lorenz.ipynb` double-click, but not the preceding folder open, because `matchSelectorInteraction() `only handles `page.click(selector)` and one-level `page.locator(selector).click() `shapes.
Could we add support and tests for locator-producing chains such as page.locator(...).getByText(...).dblclick()


Some reports are misleading because the test is not opening a file directly; it is testing context-menu behavior, `Open With`, or selection.


1. Context-menu availability / screenshots

Examples:

- [folder context menu](https://github.com/jupyterlab/jupyterlab/blob/6f058ba4d5348b5e4570b44f4639cfc2037ef7f9/galata/test/jupyterlab/contextmenu.test.ts#L45-L58)
- [file context menu](https://github.com/jupyterlab/jupyterlab/blob/6f058ba4d5348b5e4570b44f4639cfc2037ef7f9/galata/test/jupyterlab/contextmenu.test.ts#L61-L73)
- [running notebook context menu](https://github.com/jupyterlab/jupyterlab/blob/6f058ba4d5348b5e4570b44f4639cfc2037ef7f9/galata/test/jupyterlab/contextmenu.test.ts#L76-L94)

These right-clicks are intentionally opening the context menu. Suggesting `page.filebrowser.open(path)` is not equivalent because it would open the file/folder instead of leaving the context menu visible.

2. `Open With` flows

Examples:

- [`Open With > Notebook (no kernel)`](https://github.com/jupyterlab/jupyterlab/blob/6f058ba4d5348b5e4570b44f4639cfc2037ef7f9/galata/test/jupyterlab/notebook-no-kernel.test.ts#L16-L24)
- [`Open With > Editor`](https://github.com/jupyterlab/jupyterlab/blob/6f058ba4d5348b5e4570b44f4639cfc2037ef7f9/galata/test/jupyterlab/notebook-trust.test.ts#L45-L50)

If the test just wants the resulting document, these can use more specific helpers like `page.notebook.open(name, { noKernel: true })` or `page.filebrowser.open(name, 'Editor')`. But if the test is validating the menu path itself, the raw context-menu interaction should remain valid.

3. Selection / multi-selection

Example:

- [select one notebook, extend selection, then open context menu](https://github.com/jupyterlab/jupyterlab/blob/6f058ba4d5348b5e4570b44f4639cfc2037ef7f9/galata/test/jupyterlab/notebook-no-kernel.test.ts#L86-L100)

The first click here is selection, not opening. Replacing it with `page.filebrowser.open(...)` changes the test behavior.

### This one is Non-blocking/Optional

The matcher currently handles direct calls like `page.click(selector)` and one-level chains like `page.locator(selector).click()`:

[`playwright-selectors.ts#L88-L124`](https://github.com/jupyterlab/eslint-plugin/blob/2abd0b1530e79655c25c6c9243131252ca470557/src/utils/playwright-selectors.ts#L88-L124)

It does not track aliases such as:

```ts
const item = page.locator('.jp-DirListing-item:has-text("foo")');
await item.dblclick();
```
At the `await item.dblclick()` call, the rule no longer sees the original selector string, so this shape is not reported.
Alias-style locators do exist in downstream Galata tests, for example:
[filebrowser.test.ts#L33-L36](https://github.com/jupyterlab/jupyterlab/blob/6f058ba4d5348b5e4570b44f4639cfc2037ef7f9/galata/test/jupyterlab/filebrowser.test.ts#L33-L36)
```ts
const fileItem = page.locator(`.jp-DirListing-item:has-text("${fileName}")`);
await fileItem.waitFor({ state: 'visible' });

await fileItem.dragTo(dir1Breadcrumb);
```
[terminal.test.ts#L302-L307](https://github.com/jupyterlab/jupyterlab/blob/6f058ba4d5348b5e4570b44f4639cfc2037ef7f9/galata/test/jupyterlab/terminal.test.ts#L302-L307)
```ts
const folderLocator = page.locator(
  `.jp-DirListing-item:has-text("${folderName}")`
);
await folderLocator.waitFor({ state: 'visible' });
await folderLocator.click({ button: 'right' });
```
But none of it in the codebase currently has the above-mentioned shape, so it could be something to take a look at, but I think it would make it complicated, so we should probably avoid it here. It would be reasonable to document alias tracking as unsupported for now, or add simple const alias tracking later if broader coverage is desired.

Darshan808:
Thanks for the review @MUFFANUJ. I addressed most the things pointed above.

> It does not track aliases such as:
> ```js
> const item = page.locator('.jp-DirListing-item:has-text("foo")');
> await item.dblclick();
> ```

Yes, this is a known limitation and it would be quite tricky to implement this. For now I documented this limitation as you suggested.

I went through all the flagged cases and I don't see any false ones.

One more limitation is a template like ```page.dblclick(`text=${name}`)``` isn't caught, because the static part has no word character.
Example: [workspace.test.ts#L93](https://github.com/jupyterlab/jupyterlab/blob/70f642e856103473c6e90e61acf7dcbb5f1b3c84/galata/test/jupyterlab/workspace.test.ts#L93
)

We could address this with a heuristic that treats every:
```js
page.dblclick(`text=<anything>`)
```
as a file-opening action. This is true for the current JupyterLab codebase and would also simplify the logic.

However, something like:
```js
page.dblclick('text=Close >> button.jp-Button')
```
is also valid (though not used anywhere and most probably won't be) and shouldn't be flagged.

WDYT about using this heuristic ?

Darshan808:
> We could address this with a heuristic that treats every:
> ```
> page.dblclick(`text=<anything>`)
> ```
> as a file-opening action.

I don't think we should use this heuristic here.

krassowski (approved):
Thank you! I think it is better to err on the side of only flagging when we are certain so let's not expand this further - if users get annoyed they will disable the rule.

Issue #74, opened 2026-07-07
New rules: preventing `IDisposable` and `ISignal` memory leaks


krassowski:
Maybe:

- `require-disposable-ownership` - "every IDisposable created or returned must be owned, disposed, or returned."
- `require-signal-cleanup` - "every ISignal.connect() should either produce a managed IDisposable or have a corresponding disconnect()."

## Motivation

JupyterLab has a number of project-specific lifecycle conventions built around Lumino, particularly `IDisposable`, `DisposableSet`, `DisposableDelegate`, and `ISignal`. These are difficult for TypeScript to verify and are not covered by generic ESLint rules.

As a result, some classes of regressions are only discovered through testing or code review:

* resources are allocated but never disposed;
* signal connections outlive their owner;
* disposable objects are created and immediately forgotten;
* signal callbacks accidentally use the wrong `this`.

These patterns are usually mistakes, have relatively few legitimate exceptions, and should be amenable to project-specific lint rules with a high signal-to-noise ratio.

### 1. Detect disposable objects that are never owned

Flag creation of disposable objects whose lifetime is not managed.

**Bad**

```ts
new DisposableDelegate(() => {
  model.dispose();
});
```

```ts
const disposable = new DisposableDelegate(() => {
  cleanup();
});
console.log(disposable)
// never disposed or transferred
```

**Good**

```ts
this._disposables.add(
  new DisposableDelegate(() => {
    cleanup();
  })
);
```

```ts
this._disposables.add(
  signal.connect(this._onChanged, this)
);
```

```ts
return new DisposableDelegate(() => {
  cleanup();
});
```

The rule should recognize common ownership patterns such as:

* adding to a `DisposableSet`;
* returning the disposable;
* assigning to a field that is disposed elsewhere;
* immediately disposing.

---

### 2. Require ownership transfer for `IDisposable`

When a function returns an `IDisposable`, encourage the caller to either:

* store it;
* add it to a `DisposableSet`;
* return it;
* dispose it.

**Bad**

```ts
createDisposable();
```

**Good**

```ts
this._disposables.add(createDisposable());
```

```ts
const disposable = createDisposable();
...
disposable.dispose();
```

---

### 3. Encourage signal connection ownership

Connections to `ISignal` should normally be owned by something that is later disposed.

**Bad**

```ts
model.changed.connect(this._onChanged, this);
```

if there is no corresponding lifetime management.

**Good**

```ts
this._disposables.add(
[...]

PR #77, opened 2026-07-11
Add disposable ownership lint rules

MUFFANUJ:
Implements Rule 1 + Rule 2 from #74 

Adds `require-disposable-ownership` and `require-disposable-transfer` to catch newly created or returned `IDisposable` values whose ownership is dropped. The implementation recognizes Lumino disposable constructors, disposable-returning calls, `DisposableSet` / `ObservableDisposableSet`, direct `DisposableSet.from([...])` items, typed or known disposable collections, `AttachedProperty.set(...)`, returns, field assignments, class field initializers, immediate `.dispose()`, and configured ownership helper names. It also adds focused behavioral tests and docs for both rules.

Some of the known Limitations
`.set(...)` is treated as ownership only when the receiver is typed as `AttachedProperty` or matches the known `disposablesProperty` pattern; generic map-like calls such as `values.set(widget, disposables)` is not assumed safe. The rules verify that ownership is not immediately lost, but they do not prove that the owning object later disposes of stored fields. `AttachedProperty.set(...)` is accepted as ownership transfer, but later lifecycle cleanup is not verified. Control-flow analysis is conservative: cleanup in finally, exhaustive branches where all paths dispose, and alias disposal like `const alias = disposable; alias.dispose()` are not fully modeled. Catch-only disposal remains unsafe because it only runs on error paths.

### DIFF
--- a/website/docs/rules/require-disposable-transfer.md
+++ b/website/docs/rules/require-disposable-transfer.md
@@ -26,7 +26,13 @@ It accepts common ownership patterns:
 - Calling `.dispose()` immediately
 - Storing it in a variable that is later added, returned, assigned to a field,
   or disposed
-- Passing it to a configured ownership helper function
+- Passing it to a configured ownership helper function or default ownership
+  sink such as `add`, `addItem`, or `addWidget`
+
+By default, the rule does not report common borrowed-reference or
+registration-return calls such as `get`, `find`, `insertCell`,
+`contextMenuWidget`, `add`, `addCommand`, `addItem`, `registerStatusItem`,
+`addKeyBinding`, `addGroup`, `register`, `transform`, and `add*Factory`.
 
 ## Incorrect
 
@@ -63,14 +69,25 @@ disposables.dispose();
 
 ### `ownershipFunctionNames`
 
-Additional function or method names that take ownership of disposable arguments.
+Function or method names that take ownership of disposable arguments. The
+default list is `add`, `addFactory`, `addItem`, `addModelFactory`, `addWidget`,
+and `addWidgetFactory`. If provided, this list replaces the default. Set this
+option to `[]` to require stricter typed ownership checks.
+
+### `ignoredReturnFunctionNames`
+
+Function or method names whose disposable return value should be treated as
+borrowed or owned by a registration/session API. If provided, this list
+replaces the default. Set this option to `[]` to report ignored registration
+return values.
 
 ```json
 {
   "jupyter/require-disposable-transfer": [
     "warn",
     {
-      "ownershipFunctionNames": ["ownDisposable", "registerDisposable"]
+      "ownershipFunctionNames": ["ownDisposable", "registerDisposable"],
+      "ignoredReturnFunctionNames": ["addCommand", "get", "find"]
     }
   ]
 }
--- a/website/docs/rules/require-disposable-transfer.md
+++ b/website/docs/rules/require-disposable-transfer.md
@@ -10,10 +10,13 @@ Ignoring the returned value usually means the cleanup path has been lost.
 
 ## Rule details
 
-This rule checks call expressions whose return type is compatible with
-`IDisposable` or `IObservableDisposable` when TypeScript type information is
-available. It also recognizes the known Lumino factories
-`DisposableSet.from(...)` and `ObservableDisposableSet.from(...)`.
+This rule checks factory-like call expressions such as `create*`, `make*`,
+`build*`, and `new*` whose return type is compatible with `IDisposable` or
+`IObservableDisposable` when TypeScript type information is available. It also
+recognizes the known Lumino factories `DisposableSet.from(...)` and
+`ObservableDisposableSet.from(...)`. It ignores disposable values created
+directly inside a typed Jupyter plugin `activate` function, where services
+commonly live for the application lifetime.
 
 It accepts common ownership patterns:
 
@@ -38,10 +41,11 @@ fluent-initializer, or registration-return calls such as `get`, `find`,
 `getCurrent`, `contextForWidget`, `insertCell`, `insertTab`,
 `initializeState`, `contextMenuWidget`, `add`, `addCommand`, `addFileType`,
 `addItem`, `addWidgetExtension`, `addToolbarButtonClass`,
-`addCommandToolbarButtonClass`, `addTab`, `open`, `openInspector`,
+`addCommandToolbarButtonClass`, `addTab`, `delete`, `open`, `openInspector`,
 `openOrReveal`, `findWidget`, `_findWidgetByID`, `widgetAt`,
 `widgetRenderer`, `pop`, `shift`, `registerItem`, `registerStatusItem`,
-`addKeyBinding`, `addGroup`, `register`, `transform`, and `add*Factory`.
+`addKeyBinding`, `addGroup`, `register`, `set`, `transform`, and
+`add*Factory`.
 
 ## Incorrect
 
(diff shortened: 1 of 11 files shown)


Darshan808:
Thanks @MUFFANUJ. Will take a closer look soon.

FYI to test this rule in downstream projects, it should also be added to [eslint.downstream.config.mjs](https://github.com/jupyterlab/eslint-plugin/blob/main/eslint.downstream.config.mjs).

krassowski:
Taking a look at a JupyterLab downstream result:

1. There are too many errors to adopt this rule outright (1367 errors across 179 files)
2. 70% errors are in `extension` packages and only 30% core; most things in extension packages would be of low importance as these are in `activate()` body method which runs only once at registration and the deactivation pattern is not adopted at all (yet)

I used Claude to help go through the 1300+ errors; we see roughly four categories:

Bucket | Total | Ext | Core | What it is
-- | -- | -- | -- | --
A. Registration-API result | 573 (42%) | 564 | 9 | addCommand(414), addItem(80), add(20), registerStatusItem(13), addWidgetFactory(11), transform(9), register(8)…
C. Named local variable | 580 (42%) | 335 | 245 | current(143), widget(45), context(21), factory(20), cell(20), model(17), content(15)…
D. Anonymous created here | 128 (9%) | 27 | 101 | ownership rule, no variable name
B. Other call result | 86 (6%) | 32 | 54 | get(14), insertCell(11), contextMenuWidget(8)…

- Bucket A (573) are not false positives, but low-value by design. addCommand/addItem/registry add* return app-/session-lifetime registrations. In extension activate() they are almost never disposed intentionally - disposing only matters once we support deactivation. 564 of 573 are in extensions.
- Bucket B (86) - get(), insertCell(), contextMenuWidget() mostly return borrowed/owned-elsewhere objects. These are mostly false positives for the transfer rule.
- Bucket C (580) has lots of false positives; these are largely instances where access to a disposable is treated incorrectly as requiring onward transfer, e.g. [here](https://github.com/jupyterlab/jupyterlab/blob/3e69b8adb305aa482757d7d4fd4cb96267ac0a22/packages/application/src/shell.ts#L874-L878)
- Bucked D (128) has some false positives due to lack of `||` and `??` chain handling 

The next steps:
- a) Don't flag borrowed references: a variable assigned from a getter / property / index / find() (not from new or a factory). This should get rid of current(143) + cell(20) + most context/widget/model/content/get() ≈ 300+ false positives in one change.
- b) Recognize ownership-transfer sinks: shell.add, *.addWidget, tracker.add, *.addItem, DisposableSet.add, and arguments passed to super(...) / options objects.
- c) Track field assignment through ||/?? and property setters (fixes the frontend.ts/lab.ts/windowedlist FPs).
- d) Add an allow-list for the registration methods (addCommand, addItem, registerStatusItem, addKeyBinding, addGroup, register, transform, add*Factory)
 
If (a-d) still leaves us with a high number of results in `-extension` packages, we can disable these rules there for now or make them only warnings.

krassowski:
My thinking here is: let's make a rule which catches some obvious issues, but is easy to adopt, otherwise people will just turn it off in extensions. We should minimize false positives, even if it means we will introduce some false negatives. As long as there is a reasonable chunk of things it catches by default, we made an improvement, and as long as there are configuration options that allow to make the rule more strict we are not sacrificing correctness, just shipping with less annoying defaults. In that case two settings could be:
- ownership-transfer sinks (e.g. `DisposableSet.add`)
- allow list for registration methods (e.g. `addCommand`)

Darshan808:
Taking a quick look at downstream CI:

**Jupyterlite:**
1. 
```
/home/runner/work/eslint-plugin/eslint-plugin/jupyterlite/packages/apputils-extension/src/index.tsx
Error:   99:21  error  Disposable "tracker" must be owned, returned, assigned to a field, or disposed  jupyter/require-disposable-ownership
```
[This one](https://github.com/jupyterlite/jupyterlite/blob/668756d22b4bd055fd9075406037f02870ef241f/packages/apputils-extension/src/index.tsx#L99-L101) (and many more from notebook and lab CI) is from the `activate` function and as mentioned above they are almost never disposed intentionally. Let's not flag these.

2.

```
/home/runner/work/eslint-plugin/eslint-plugin/jupyterlite/packages/services/src/kernel/client.ts
Error:   279:5  error  IDisposable returned from "set" must be stored, returned, added to a DisposableSet, or disposed     jupyter/require-disposable-transfer
Error:   308:7  error  IDisposable returned from "delete" must be stored, returned, added to a DisposableSet, or disposed  jupyter/require-disposable-transfer
```
I believe these are flagged because `set` and `delete` on `ObservableMap` return an `IDisposable`

The [delete here is](https://github.com/jupyterlite/jupyterlite/blob/668756d22b4bd055fd9075406037f02870ef241f/packages/services/src/kernel/client.ts#L306-L312) inside `kernel.disposed.connect(...)` which is a pure false positive. This handler only runs because the kernel is already being disposed.

For [set here](https://github.com/jupyterlite/jupyterlite/blob/668756d22b4bd055fd9075406037f02870ef241f/packages/services/src/kernel/client.ts#L279), I think the intuition here is that since `set` returns the old value for the key, it must be handled. But since `startNew` uses a fresh/unique kernel id, so the old value is undefined.

As suggested above by Mike let's not flag borrowed references and only flag functions that transfer ownership on a newly created reference (using new or a factory like `createDisposable()`)


**Notebook**
Most of the flagged errors are from `activate` function.

Only one flagged case from core package is:
```
/home/runner/work/eslint-plugin/eslint-plugin/notebook/packages/application/src/panelhandler.ts
Error:   190:20  error  IDisposable "widget" must be returned, added to a DisposableSet, assigned to a field, or disposed  jupyter/require-disposable-transfer
```

```ts
const widget = this._findWidgetByID(id);
```
This one's also borrowed references.

After fixing these cases, I think we can reduce false positives even more.

krassowski:
^picked up a few conflicts too

krassowski:
Are the copilot suggestions traceable to real test cases we could add and could occur in the codebase? If so, then probably worth addressing.

MUFFANUJ:
Maybe make severity of test as 'warn' here?

website/docs/rules/require-disposable-transfer.md
```diff
@@ -0,0 +1,108 @@
...
+By default, the rule does not report common borrowed-reference,
+fluent-initializer, or registration-return calls such as `get`, `find`,
+`getCurrent`, `contextForWidget`, `insertCell`, `insertTab`,
+`initializeState`, `contextMenuWidget`, `add`, `addCommand`, `addFileType`,
+`addItem`, `addWidgetExtension`, `addToolbarButtonClass`,
+`addCommandToolbarButtonClass`, `addTab`, `delete`, `open`, `openInspector`,
+`openOrReveal`, `findWidget`, `_findWidgetByID`, `widgetAt`,
+`widgetRenderer`, `pop`, `shift`, `registerItem`, `registerStatusItem`,
+`addKeyBinding`, `addGroup`, `register`, `set`, `transform`, and
+`add*Factory`.
```
krassowski, line 48: This is likely to diverge from the source code and probably too extensive of a list for docs; can we link to a symbol instead? Not easy since we don't have code docs, but maybe "for full list search for `DEFAULT_IGNORED_RETURN_FUNCTION_NAMES` and link to GitHub search for this repo?

website/docs/rules/require-disposable-transfer.md
```diff
@@ -0,0 +1,108 @@
...
+default list is `add`, `addCell`, `addFactory`, `addItem`, `addModelFactory`,
+`addSibling`, `addMenu`, `addWidget`, `addWidgetFactory`, `insertItem`,
+`insertWidget`, and `registerStatusItem`. If provided, this list replaces the
+default. Set this option to `[]` to require stricter typed ownership checks.
+
+### `ignoredReturnFunctionNames`
+
+Function or method names whose disposable return value should be treated as
+borrowed or owned by a registration/session API. If provided, this list
+replaces the default and opts into stricter return checking for non-factory
```
krassowski, line 95: The default ignore list is quite large; I wonder if instead it should "add" to the default list (and then in future if needed we could have another setting for skipping default ignores)

krassowski (approved):
I think we can merge and iterate on it. The "replace" semantics for ignores are the biggest issue IMO but we can change it later.

PR #78, opened 2026-07-13
Add `require-signal-cleanup`, `require-signal-this-arg` and `prefer-signal-this-arg`

Darshan808:
## References #74 (Rules 3 and 4)

This PR adds three new rules.

| Rule | Recommended level | Catches |
| --- | --- | --- |
| `require-signal-cleanup` | `warn` | Connections to a proven long-lived application service, made with `this` as receiver, from a non-derived class that has a disposal protocol but shows no cleanup evidence |
| `require-signal-this-arg` | `error` | Bare method callbacks that break at runtime without a `thisArg` |
| `prefer-signal-this-arg` | `warn` | Other connections without a `thisArg`, where the sender is proven to outlive the receiver and the class already cleans up by receiver (unclearable by `Signal.clearData(this)`) |

Both `require-signal-cleanup` and `prefer-signal-this-arg` require type information. Without a type checker configured, neither rule reports anything — this is intentional; the alternative is name-based guessing, which produces false positives.

---

## Rule 1: `require-signal-cleanup`

Flag `signal.connect(callback, this)` when the sender is a **proven application-lifetime service** and the enclosing class - which has a teardown point - shows no cleanup evidence of any kind.

The rule does **not** ask "does this class clean up?" in isolation, a class that constructs and disposes its own sender shouldn't need to. Instead it asks the governing question behind all these rules:

> **Does the sender outlive the receiver?**

Only when the sender is proven to be a long-lived, injected service (via a curated allowlist - `CommandRegistry`, `IDebugger`, `IDocumentManager`, `ILSPConnection`, `ILabShell`, `ILanguageServerManager`, `IRenderMimeRegistry`, `ISettingRegistry`, `IShell`, `IStateDB`, `IThemeManager`, `ServiceManager`, configurable via `longLivedTypes`) does the rule proceed to check for cleanup.

### Decision pipeline

```text
1.  Is it <expr>.connect(...)?                          no → ignore
2.  Does it have ≥ 2 arguments and is arg 1 `this`?     no → ignore (Rule 2/3's domain)
3.  Is the call lexically inside a class?               no → ignore (app-lifetime code)
4.  Does that class have an `extends` clause?           yes → ignore (may inherit dispose)
5.  Does the class have a disposal protocol
    (implements *Disposable, or a dispose/
    isDisposed member)?                                 no → ignore (nowhere to put a fix)
6.  Is the signal named `disposed`?                     yes → ignore (self-terminating)
7.  Does the class body contain cleanup evidence?       yes → ignore
8.  Does the receiver's TS type resolve to
    something that is NOT `ISignal`/`Signal`?           yes → ignore
9.  Does the sender resolve to a proven
    long-lived service?                                 no → ignore
10. ⚠ Report `serviceOutlivesReceiver`
```

Cleanup evidence (any one silences the class):

- A `Signal.clearData(...)`, `Signal.disconnectReceiver(...)`, `Signal.disconnectAll(...)`, `Signal.disconnectSender(...)`, or `Signal.disconnectBetween(...)` call (including renamed imports)
- Any `.disconnect(...)` call
- A method named in the `additionalCleanupMethods` option

---

## Rule 2: `require-signal-this-arg` (`error`)

Flag `signal.connect(this.method)`, a `connect` call with exactly one argument when `method` is a regular method of the enclosing class whose body actually uses `this`.

The runtime bug is that Lumino invokes the slot with the registered `thisArg`. If no `thisArg` is provided, `this` inside the callback is `undefined` (in strict mode), so the first `this.anything` access throws long after the code ships.

This rule reports **only** the shapes that break at runtime; every other missing-`thisArg` case belongs to Rule 3.

### Decision pipeline

```text
1. Is it <expr>.connect(<one argument>)?               no → ignore
2. Is the argument a bare `this.x` / `this.#x`?        no → ignore
3. Is the call inside a class?                         no → ignore
4. Does `x` resolve to a member of that class
   (matching static-ness and `#` privacy)?             no → ignore (inherited?)
5. Is the member a regular method
   (not an arrow property, getter, or setter)?         no → ignore
6. Does the method body use `this`
   (arrow-transparent, function-opaque)?               no → ignore
7. Receiver type:
      • `not-signal` → ignore
      • `unknown`    → require a signal-like name
                        fails → ignore
8. ⚠ Report `missingThisArg` with a suggestion to add `, this`
```

This rule doesn't require type information to fire at all — it keeps a name-based fallback (`/(signal|changed|disposed)$/i`) for the `unknown` case, since its trigger is a provable runtime bug rather than a lifetime judgement.

---

## Rule 3: `prefer-signal-this-arg` (`warn`)

Inside a class, flag a one-argument `signal.connect(callback)` **only when the sender is proven to outlive the receiver** and the class already cleans up by receiver.

These callbacks work at runtime, but the connection is registered with no receiver. `signal.disconnect(callback, thisArg)` and `Signal.clearData(thisArg)` match connections **by receiver**, so a connection registered without a `thisArg` cannot be removed by `Signal.clearData(this)`. That only matters if the sender is still around to fire the callback - hence the same sender-lifetime proof as Rule 1 (an allowlisted service, or a model reached through a `Widget` view), plus a check that adding `, this` would actually be picked up by something (`Signal.clearData(this)`, a matching `.disconnect(cb, this)`, or an inherited `Widget.dispose()`).

Passing a `thisArg` is safe for callback binding in every shape, but **not** safe for cleanup matching: Lumino's `findConnection` matches the exact triple `(signal, slot, thisArg)`, so if the same callback is already removed elsewhere via a one-argument `disconnect(cb)`, adding `, this` here would break that teardown. The rule checks the whole file for such a match before reporting.

### Decision pipeline

```text
1.  Is it <expr>.connect(<one argument>)?               no → ignore
2.  Is the call inside a class?                         no → ignore
3.  Is the signal named `disposed`?                     yes → ignore (self-terminating)
4.  Is it Rule 2's case (bare `this.x` method
    that uses `this`)?                                  yes → ignore (partition)
5.  Is the callback shape 'opaque'
    (a call result, a conditional, etc.)?                yes → ignore
6.  Is there a matching 1-arg `disconnect(cb)`
    anywhere in the file?                                yes → ignore (would break it)
7.  Does the class clean up by receiver
    (Signal.clearData(this)/disconnect(cb, this)/etc.),
    or does it extend a Lumino `Widget`?                 no → ignore (fix wouldn't help)
8.  Receiver type:
       • `not-signal` → ignore
       • `unknown`    → require a signal-like name
                         fails → ignore
9.  Does the sender resolve to a proven long-lived
    service or a model behind a `Widget` view?           no → ignore
10. ⚠ Report `preferThisArg` (or `boundCallback` for `.bind(this)` callbacks)
```


website/docs/rules/require-signal-this-arg.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Why
+
+Lumino's `ISignal.connect(callback, thisArg)` invokes the callback with `thisArg` as its receiver. When a class method is passed as a bare function reference, `signal.connect(this._onChanged)`, nothing binds `this` inside the callback to the instance, so any `this.` access in the method body throws or reads the wrong object when the signal fires.
+
+A matching `thisArg` also matters for cleanup: `signal.disconnect(callback, thisArg)` and `Signal.clearData(thisArg)` only remove connections whose receiver matches.
+
+## Rule details
+
+The rule reports `signal.connect(this.method)` calls - exactly one argument, where the argument is a reference to a member of the enclosing class, when the referenced method's body actually uses `this`. Usage inside nested arrow functions counts as arrows inherit `this` lexically; usage only inside nested regular `function`s does not.
```
Darshan808, line 13: As mentioned here and in issue too, `this` usage inside nested arrow functions work:
> **Good**
> ```js
> model.changed.connect(sender => {
>   this._onChanged(sender);
> });
> ```

But the [docs here](https://jupyterlab.readthedocs.io/en/stable/developer/patterns.html#signals) says:
> Wherever possible a signal connection should be made with the pattern .connect(this._onFoo, this). Providing the this context enables the connection to be properly cleared by Signal.clearData(this).

So should we also recommend to pas `thisArg` when the first argument is arrow function even though there might not be direct runtime bugs (but might be memory leakage) ?

website/docs/rules/require-signal-cleanup.md
```diff
@@ -0,0 +1,109 @@
...
+- any `.disconnect(...)` call (covers `dispose()` teardown as well as disconnect-before-reconnect idioms such as a `stopObserving()` helper)
+- consuming the return value of a `.connect(...)` call (for example adding it to a disposable collection)
+- wiring cleanup through a `disposed`-style signal (`x.disposed.connect(...)`) or delegating via a `this.dispose()` call
+
+The rule is deliberately conservative and skips:
+
+- classes with an `extends` clause — a base class such as Lumino's `Widget` already calls `Signal.clearData(this)` in its inherited `dispose()`
+- `.connect()` calls outside any class (module scope, plugin `activate()` functions) — these are typically app-lifetime connections with nothing to leak into
+- calls where the second argument is not `this` — another object's lifecycle cannot be traced from here
+- single-argument `.connect(callback)` calls — see [require-signal-this-arg](../require-signal-this-arg)
```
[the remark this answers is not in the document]
Darshan808: Yes. But when I tried with `(./require-signal-this-arg)`, the URL went to `https://eslint-plugin/en//rules/require-signal-cleanup/require-signal-this-arg` and only on using  `(../require-signal-this-arg)`, it went to the correct URL.
Can be verified from **readthedocs** CI action below.

website/docs/rules/require-signal-this-arg.md
krassowski: Can we distinguish this at warning/error levels? Not passing `this` would be only a warning, unless it is used without an `arrow` then it would be an error. In practice this means two rules, we can switch both to `error` later if warranted. Any downsides?

Darshan808: > In practice this means two rules

I think this is the only solution here to distinguish this at warning/error levels.
So now `require-signal-this-arg` is for reporting errors when `this` isn't passed to class method that references `this`. (Runtime Bug)
and `prefer-signal-this-arg` is for reporting warnings when `this` isn't passed to class method that don't reference this or arrow functions. (Memory leakage Bug)

Darshan808:
#### From the jupyterlab downstream test

| Rule | Count |
|---|---|
| `jupyter/prefer-signal-this-arg` | 88 |
| `jupyter/require-signal-cleanup` | 9 |
| `jupyter/require-signal-this-arg` | 0 |
| **Total** | **97** |

`prefer-signal-this-arg` are mostly from not passing thisArg with arrow functions.
`require-signal-this-arg` didn't flag any which means there isn't any case where `this` is being used in callback and is not being supplied when connecting with signal.
`require-signal-cleanup` flagged 9 cases where we're connecting the callback with signal but there isn't any cleanup evidence in the class. (I believe some of these are classes which live throughout the application lifetime like  `ThemeManager` class which is being flagged)

I think we should reduce `prefer-signal-this-arg` as `warning` instead of `error` in downstream test for now as I think we would it as `warning` in lab too.

krassowski:
> I think we should reduce prefer-signal-this-arg as warning instead of error in downstream test for now as I think we would it as warning in lab too.

I agree that it's probably fine to add as a warning for lab for now, but if there are only a handful of instances in Notebook and Lite maybe it's worth enforcing it there?

> require-signal-cleanup flagged 9 cases where we're connecting the callback with signal but there isn't any cleanup evidence in the class

Sounds good.

^picked up a few conflicts

tests/fixtures/signaling.d.ts
```diff
@@ -0,0 +1,37 @@
...
+  connect(slot: Slot<SENDER, ARGS>, thisArg?: unknown): boolean;
+  disconnect(slot: Slot<SENDER, ARGS>, thisArg?: unknown): boolean;
+}
+
+export declare class Signal<SENDER, ARGS> implements ISignal<SENDER, ARGS> {
+  constructor(sender: SENDER);
+  connect(slot: Slot<SENDER, ARGS>, thisArg?: unknown): boolean;
+  disconnect(slot: Slot<SENDER, ARGS>, thisArg?: unknown): boolean;
+  emit(args: ARGS): void;
+}
```
krassowski: Hmm why not pull `@lumino/signalling` as dev dependency for tests here?

website/docs/rules/require-signal-cleanup.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Rule details
+
+The rule inspects every `signal.connect(callback, this)` call - two arguments, with `this` as the receiver, made inside a class, and reports it when the class shows **no cleanup evidence anywhere in its body**. Any one of the following silences the whole class:
+
+- a call to `Signal.clearData(...)`, `Signal.disconnectReceiver(...)`, `Signal.disconnectAll(...)`, `Signal.disconnectSender(...)`, or `Signal.disconnectBetween(...)` (including via a renamed import of `Signal` from `@lumino/signaling`)
+- any `.disconnect(...)` call (covers `dispose()` teardown as well as disconnect-before-reconnect idioms such as a `stopObserving()` helper)
+- consuming the return value of a `.connect(...)` call (for example adding it to a disposable collection)
```
krassowski: That's incorrect and traces down to a bad example I gave in #74. I tried to marry two issues together and gave an incorrect:

```ts
this._disposables.add(
  signal.connect(this._onChanged, this)
);
```

That does not make sense as signal returns a `boolean` not a `IDisposable` (https://lumino.readthedocs.io/en/latest/api/interfaces/signaling.ISignal.html#connect.connect-1).

website/docs/rules/prefer-signal-this-arg.md
```diff
@@ -0,0 +1,86 @@
+# `prefer-signal-this-arg`
+
+Prefer passing a `thisArg` when connecting to a Lumino signal so the connection can be cleaned up.
+
+## Why
+
+The JupyterLab documentation recommends making signal connections with the pattern `.connect(this._onFoo, this)` wherever possible. The `thisArg` is stored as the connection's **receiver**, and both `signal.disconnect(callback, thisArg)` and `Signal.clearData(thisArg)` match connections by receiver. A connection registered without a `thisArg` has no receiver, so `Signal.clearData(this)` in a `dispose()` method silently fails to remove it — the connection can outlive the object and leak, even when the callback itself works fine at runtime.
```
krassowski: Should we add a link? https://jupyterlab.readthedocs.io/en/latest/developer/patterns.html#signals (and maybe a back link to here).

website/docs/rules/prefer-signal-this-arg.md
```diff
@@ -0,0 +1,86 @@
+# `prefer-signal-this-arg`
```
krassowski, line 1: There is a major issue with this rule that I had not realised before:
- the `.connect(callback, this)` must be accompanied by `signal.disconnect(callback, this)` or `Signal.clearData(this)` - so far ok
- adding `this` to an existing `.connect(callback)` so making it into `.connect(callback, this)` will break existing `signal.disconnect(callback)` because the matching is exact in the private [`findConnection`](https://github.com/jupyterlab/lumino/blob/761a1d0ff66752cc8ac7463b63dc261592ab4961/packages/signaling/src/index.ts#L691-L704) function.

src/utils/signals.ts
```diff
@@ -0,0 +1,540 @@
...
+  let found = false;
+  walkFrom(classNode.body, node => {
+    if (node.type === 'ClassDeclaration' || node.type === 'ClassExpression') {
+      return 'skip-children';
+    }
+    if (node.type === 'CallExpression') {
+      if (
+        isSignalNamespaceCleanupCall(node, signalLocalNames) ||
+        isDisconnectCall(node, extraMethodNames) ||
+        isThisDisposeCall(node) ||
```
krassowski: This is used for `require-signal-cleanup` but:
- `dispose(): void {}` cleans up nothing
- one stray `this.dispose()` call, silences the class
- inherited `dispose()` is not detected?

website/docs/rules/prefer-signal-this-arg.md
```diff
@@ -0,0 +1,86 @@
...
+
+Prefer passing a `thisArg` when connecting to a Lumino signal so the connection can be cleaned up.
```
krassowski: We need change this rule to:


```suggestion
Pass a `thisArg` when connecting to a Lumino signal that is cleaned up with `Signal.clearData(this)` or with `signal.disconnect(callback, thisArg)`.
```

krassowski: This requires type inference so when we inherit from say lumino's `Widget` which has `Signal.clearData(this)` in its [`dispose()`](https://github.com/jupyterlab/lumino/blob/761a1d0ff66752cc8ac7463b63dc261592ab4961/packages/widgets/src/widget.ts#L55-L82) method this rule should flag.

src/utils/signals.ts
krassowski: This requires type inference so when we inherit from say lumino's `Widget` which has `Signal.clearData(this)` in its [`dispose()`](https://github.com/jupyterlab/lumino/blob/761a1d0ff66752cc8ac7463b63dc261592ab4961/packages/widgets/src/widget.ts#L55-L82) method this rule should NOT flag.

website/docs/rules/require-signal-cleanup.md
```diff
@@ -0,0 +1,109 @@
+# `require-signal-cleanup`
```
krassowski, line 1: This would ideally be reported once per class, not once per connect. `shell.ts` got 6 identical messages for one class.

website/docs/rules/prefer-signal-this-arg.md
```diff
@@ -0,0 +1,86 @@
...
+
+Inside a class, the rule reports one-argument `signal.connect(callback)` calls for every callback shape that does not have the runtime `this` bug, including:
+
+- inline arrow functions and function expressions
+- arrow-function class properties (lexically bound, so no runtime bug — but still not clearable)
+- class methods that never reference `this`
+- members not found in the enclosing class (possibly inherited)
+- free-variable callbacks and `.bind(this)` calls
+
+A **suggestion** (editor quick-fix, not an autofix) is offered to append `, this` as the second argument. Passing a `thisArg` is valid for every callback shape: it sets the connection's receiver without changing how an arrow function binds `this`.
```
krassowski: > Passing a `thisArg` is valid for every callback shape

This is over-confident - there are cases where we don't want it. Consider

https://github.com/jupyterlab/jupyterlab/blob/5fc80a1c9c8c0d7768abb345add7e98fedc4fa30/packages/toc/src/factory.ts#L64-L101

Opus 5:


> Wrong reason 1 (the general blocker): L98/L99 already tear these down, per-widget, at exactly the right moment. Accepting the suggestion on L81 makes L98 stop matching — the connection becomes permanent.
>
> Wrong reason 2 (specific to this shape): this here is the [TableOfContentsFactory](https://github.com/jupyterlab/jupyterlab/blob/5fc80a1c9c8c0d7768abb345add7e98fedc4fa30/packages/toc/src/factory.ts#L22) — one instance registered with the ToC registry at startup, never disposed. createNew runs once per opened document, so you'd anchor every document's connection to an object nothing ever calls Signal.clearData on. The callbacks aren't even members of this (local const arrows), so the receiver expresses no real ownership. The edit trades working per-widget teardown for a permanent registration on an immortal object — strictly worse than leaving it alone.
>
> Same shape at [labicon.tsx#L472-L486](https://github.com/jupyterlab/jupyterlab/blob/5fc80a1c9c8c0d7768abb345add7e98fedc4fa30/packages/ui-components/src/icon/labicon.tsx#L472-L486) (useEffect cleanup) and [vdom.ts#L264-L270](https://github.com/jupyterlab/jupyterlab/blob/5fc80a1c9c8c0d7768abb345add7e98fedc4fa30/packages/ui-components/src/components/vdom.ts#L264-L270) (componentWillUnmount).

website/docs/rules/require-signal-cleanup.md
```diff
@@ -0,0 +1,109 @@
...
+
+The rule is deliberately conservative and skips:
+
+- classes with an `extends` clause — a base class such as Lumino's `Widget` already calls `Signal.clearData(this)` in its inherited `dispose()`
+- `.connect()` calls outside any class (module scope, plugin `activate()` functions) — these are typically app-lifetime connections with nothing to leak into
+- calls where the second argument is not `this` — another object's lifecycle cannot be traced from here
+- single-argument `.connect(callback)` calls — see [require-signal-this-arg](../require-signal-this-arg)
+
+When type information is available, receivers whose type does not resolve to Lumino's `ISignal`/`Signal` are ignored.
+## Incorrect
```
krassowski, line 37:
```suggestion

## Incorrect
```

website/docs/rules/prefer-signal-this-arg.md
krassowski: There are four false positives where applying the suggestion to add `, this` as suggested in the current form would lead to a new memory leak.

krassowski (changes requested):
It looks like `prefer-signal-this-arg` requires changes, `require-signal-cleanup` could be much better.

Darshan808:
The decision pipeline turned out to be a lot trickier to implement than expected. The original receiver-centric approach (flag a class with no visible cleanup) produced a lot of false positives in lab's codebase. Classifying every finding showed `require-signal-cleanup` was running at very low precision and `prefer-signal-this-arg` at about 40%.

Rewrote both rules to be sender-centric and presence-based instead: rather than asking "does this class clean up?", they now ask "can I *prove* this sender outlives this receiver?" and only report on a positive proof (an allowlisted long-lived service, or a model behind a `Widget` view).

After adding these stricter gates and matching only the obvious cases:

- `require-signal-cleanup` now reports **0 findings** in lab's codebase
- `prefer-signal-this-arg` now reports **15 findings, all of which look like true positives**
- `require-signal-this-arg` still flags 0 cases

krassowski (approved):
Thank you @Darshan808. This looks good to me. In addition to jupyterlab I scanned 5 extensions for better coverage.

Review with Opus 5 locally found some non-blocking suggestions, listing them below:

> Medium
> 
> - Neither rule checks that this at the connect site is the enclosing instance, so both report connects in static methods and in nested non-arrow functions. In prefer-signal-this-arg.ts:193-212 the offered suggestion then inserts a this that is the class constructor (static) or undefined (nested function), which no instance `dispose()` can ever match and which is a type error under `noImplicitThis`. Reproduced: a `static wire(registry: ISettingRegistry)` inside a `Widget` subclass is reported and offered `, this`; so is a connect inside function `inner()` in a method. require-signal-cleanup.ts:112 has the mirror case, reporting a `static connect(cb, this)` with "it will outlive this class and keep every instance alive". `isStaticContext` already exists in signals.ts:192 but is only used by `resolveUnboundThisMethodConnect`.
> - `isLongLivedService` in signal-lifetime.ts:747-771 starts the walk at `chain.prefixes[0]`, which is the this node itself when the chain is `this.x.signal`. A class whose own type name is in `longLivedTypes` therefore matches itself, and every connect to one of its own non-owned fields is reported as "…is an application-lifetime service that will outlive this class". Two of the twelve default entries are concrete class names in the ecosystem (`CommandRegistry,` `ServiceManager`), and both rule docs actively tell users to add their own long-lived class names to the option. Reproduced with a `class ServiceManager implements IDisposable` assigning `this.sessions = sessions` in its constructor: `require-signal-cleanup` reports `this.sessions`. jupyterlab's real `ServiceManager` escapes only because it builds its sub-managers with new, which the ownership pass catches first.
> 
> Low
> 
> - prefer-signal-this-arg.ts:44: the `boundCallback` message ends "Connect the method directly and pass this as the second argument" for any `.bind(...)`, including `.bind(someOtherObject)`, where following it would move the callback's receiver off that object. `getUnbindReplacement` correctly declines to emit a suggestion in that case, so only the message text is wrong.
> - signals.ts:129-150 collects Signal aliases only from named imports, so a class cleaning up via `import * as signaling from '@lumino/signaling'` and `signaling.Signal.clearData(this)` is not credited with cleanup evidence and gets reported. No occurrences downstream (0 namespace imports of `@lumino/signaling` across the scanned repos), so this is latent.
> 
>

PR #84, opened 2026-07-28
Fix pre-commit so it actually lints

### LINKED ISSUE #83: Run lint tests on pre-commit
### Description
pre-commit only runs the lint rule tests. Due to which general lint test on codebase is discovered later on CI.
Let's run lint test like `eslint` on pre-commit too.

Darshan808:
### Summary

This PR fixes the project's pre-commit configuration so the intended linting checks actually run before commits.

With the lint hook working correctly, it surfaced a number of existing formatting issues which were auto fixed and are included in this PR.


krassowski (approved):
Thank you!

PR #85, opened 2026-07-29
Add `galata-prefer-menu-helper` rule

Darshan808:
### Description
**Detects:** raw Playwright menu traversal in UI tests; suggests `page.menu` helper.

## Example
```js
// flagged
await page.click('text=File');
await page.click('.lm-Menu ul[role="menu"] >> text=New');
await page.click('#jp-mainmenu-file-new >> text=Terminal');

// suggested
await page.menu.clickMenuItem('File>New>Terminal');
```

## Messages
- `preferMenuOpen` — menu bar item
- `preferClickMenuItem` — item inside an already-open menu
- `preferMenuHelper` — menu markup with no derivable label path


Darshan808:
**38** flagged cases of `galata-prefer-menu-helper` in lab

MUFFANUJ:
Hey @Darshan808, FYI, please do reference the issue in the PRs earlier when you work on those, as it gets missed and creates duplicate work, as happened recently with #86. Thanks!

Darshan808:
Ah, sorry about that! It was still in the draft, so I hadn’t updated the description. I’ll make sure to at least add the issue reference going forward.

MUFFANUJ:
No worries, thanks!

Darshan808:
@MUFFANUJ  since you’ve already worked on this, I think it’d be better if you could give it a quick review. Thanks!

MUFFANUJ:
Thanks, this is useful, and the JupyterLab raw-selector cases are being caught well. I rechecked this against the current PR head (`ad2ae0f`) and current downstream main branches, and I think there is one downstream coverage gap before this rule is enabled for Notebook/JupyterLite UI tests.

The downstream config enables this rule for both `galata/test` and `ui-tests/test`:
https://github.com/jupyterlab/eslint-plugin/blob/ad2ae0f1d3fbf16e7a4e88ce49ef01756cdc92d9/eslint.downstream.config.mjs#L127-L144

But the rule currently only goes through `matchSelectorInteraction(...)`, and the tests explicitly treat `getByRole(...)` chains as unmatched: since there is a comment, maybe this is intentional, but should we include this also?
https://github.com/jupyterlab/eslint-plugin/blob/ad2ae0f1d3fbf16e7a4e88ce49ef01756cdc92d9/tests/jupyter-galata-prefer-menu-helper.test.ts#L42-L44

That leaves real downstream menu traversal unreported, for example:

- Notebook uses role-based top-level menu traversal:
  https://github.com/jupyter/notebook/blob/32faac4d874dabfc650f432af020799d278dc4f2/ui-tests/test/filebrowser.spec.ts#L90-L91

- JupyterLite does the same for `Settings`:
  https://github.com/jupyterlite/jupyterlite/blob/7ed8f1cca6b3790bfe84df14c39109d0a97f3262/ui-tests/test/page.spec.ts#L49-L51

There is also a Notebook-specific top-level `New` menu shape that is missed because `TOP_LEVEL_MENU_LABELS` is JupyterLab-only (`File|Edit|View|Run|Kernel|Tabs|Settings|Help`):
https://github.com/jupyterlab/eslint-plugin/blob/ad2ae0f1d3fbf16e7a4e88ce49ef01756cdc92d9/src/rules/galata-prefer-menu-helper.ts#L15

Real examples:
https://github.com/jupyter/notebook/blob/32faac4d874dabfc650f432af020799d278dc4f2/ui-tests/test/mobile.spec.ts#L52
https://github.com/jupyter/notebook/blob/32faac4d874dabfc650f432af020799d278dc4f2/ui-tests/test/smoke.spec.ts#L28

I verified this with the compiled rule: `text=File` and `.lm-Menu ... text=New` report, but `text=New`, `page.getByRole('menuitem', { name: 'File' }).click()`, and `page.getByRole('menuitem', { name: 'Settings' }).click()` report nothing.

Darshan808:
> since there is a comment, maybe this is intentional, but should we include this also?

Yes this was intentional as it was conflicting with the context menu. But I think we can make it work. I'm including it. Thanks for pointing out.

> There is also a Notebook-specific top-level New menu shape that is missed because TOP_LEVEL_MENU_LABELS is JupyterLab-only (File|Edit|View|Run|Kernel|Tabs|Settings|Help):

Yes. But I believe we are making this linter for general extensions and that `New` is very specific to Notebook. Including that would definitely flag the wrong usage in notebook's codebase but that `New` is a submenu inside `File` in jupyterlab. Extension authors' reference UI is JupyterLab, so for them a top-level `New` doesn't exist in my opinion.

Darshan808:
All 36 warnings across 8 files in lab looks like true positives. Some warnings can be suppressed in the lab codebase itself if we want to keep them as it is.

MUFFANUJ:
> Yes. But I believe we are making this linter for general extensions and that `New` is very specific to Notebook. Including that would definitely flag the wrong usage in notebook's codebase but that `New` is a submenu inside `File` in jupyterlab. Extension authors' reference UI is JupyterLab, so for them a top-level `New` doesn't exist in my opinion.

Yup Make sense

MUFFANUJ (approved):
LGTM!

krassowski:
Similar to the initial state of #105 it does not follow `const` declarations so it can produce a false positive on:

```js
const item = page.locator('.jp-DirListing-item');
await item.click({ button: 'right' });
await page.click('.lm-Menu ul[role="menu"] >> text=Rename');
```

or 

```ts
const menu = page.locator('.lm-Menu');
await menu.locator('.lm-Menu-item:has-text("Set Default Threshold")').click();
```

even though inlined it works:

```ts
await page.click('.jp-DirListing-item', { button: 'right' });
await page.click('.lm-Menu ul[role="menu"] >> text=Rename');
```

or 

```ts
await page.locator('.lm-Menu').locator('.lm-Menu-item:has-text("Set Default Threshold")').click();
```


The suppression on right-click is also too coarse as it does not account for scoping:

```ts
test('a', async ({ page }) => {
  await page.click('.lm-Menu ul[role="menu"] >> text=New');   // preferClickMenuItem
});
test('b', async ({ page }) => {
  await page.click('.jp-DirListing-item', { button: 'right' });
  await page.click('.lm-Menu ul[role="menu"] >> text=Rename');
});
test('c', async ({ page }) => {
  await page.click('.lm-Menu ul[role="menu"] >> text=Close Tab');  // not reported
});
```

Finally, this rule cannot distinguish the menus opened from menu bar vs menu opened from a toolbar, e.g. in
- https://github.com/jupyterlab/jupyterlab/blob/fcd2ffd7ba26449ed8dfc2b61dc3378deac40e83/galata/test/documentation/debugger.test.ts#L158
- https://github.com/deshaw/jupyterlab-notify/blob/c661786d52664cd8489b34c58adcf7f6bc0425cf/ui-tests/tests/helpers.ts#L74

Not sure what to do here yet. Maybe a conditional message "If this is the top-level menu use ..."

krassowski (approved):
Thanks!

Issue #71, opened 2026-06-29
Add custom ESLint rules to enforce Galata helper usage in tests


MUFFANUJ:
## Problem

JupyterLab's integration and visual tests frequently bypass Galata's higher-level helper API and use raw Playwright selectors and interactions instead. This creates several problems:

- Raw selectors break easily when UI labels or class names change
- Manual interactions skip the built-in waits and readiness checks that 
  Galata helpers provide, causing flaky tests
- The same multi-step interaction is repeated across many test files 
  instead of being expressed in one helper call
- Tests become harder to read and maintain over time

For example, opening a notebook currently looks like this in many tests:

```ts
// raw approach — brittle and missing waits
await page.dblclick('[aria-label="File Browser Section"] >> text=notebooks');
await page.dblclick('text=Data.ipynb');
```

When Galata already provides:

```ts
// helper approach — handles waits, directory traversal, tab visibility
await page.notebook.openByPath('notebooks/Data.ipynb');
```

This pattern repeats across file browser interactions (113 hits), notebook cell operations (102 hits), context menu flows (51 hits), main menu traversal (35 hits), and sidebar/activity tab selection (14 hits).

## Proposed Solution

We could initiate this by adding few custom ESLint rules to the here that detect raw Playwright patterns in test files and suggest or require the equivalent Galata helper instead and have different severity for each rule depending on its impact and necessity.

These rules should apply only to test files such as 
`galata/test/**/*.ts`. They should NOT apply to `galata/src/helpers/**` 
since that is the helper implementation itself and must contain low-level 
Playwright operations.

Few suggestion on rules from what i observed from most of visual/integration tests
---

## Rule 1: `galata/prefer-filebrowser-helper`

**Severity: Error** for normal file/directory opening flows.

Detects raw file browser interactions using selectors like 
`.jp-DirListing-item`, `.jp-BreadCrumbs-home`, or double-clicks on 
file browser text.

```ts
// Bad
await page.dblclick('[aria-label="File Browser Section"] >> text=notebooks');
await page.dblclick('text=Data.ipynb');
await page.click('.jp-BreadCrumbs-home svg');

// Good
await page.notebook.openByPath('notebooks/Data.ipynb');
await page.filebrowser.openHomeDirectory();
await page.filebrowser.open('data/bar.vl.json');
```

[...]

PR #86, opened 2026-08-04
Add Galata rule to prefer MenuHelper for main menu traversal

MUFFANUJ:
Part of : #71 

Adds `prefer-menu-helper` to detect raw Playwright main-menu traversal in Galata tests and recommend `page.menu.clickMenuItem()` / `page.menu.openLocator()` instead. The rule covers top-level text, `#jp-mainmenu-*`, Lumino menu selectors, role/text locators, chained locators, aliases, and follow-up submenu clicks, while suppressing context-menu flows such as `Open With`. It is enabled as `warn` in recommended test-file configs, added to downstream scan config as `error`, and documented with examples. Includes focused RuleTester coverage for valid helper/context-menu cases and invalid raw main-menu patterns.


MUFFANUJ:
Hey @Darshan808, this one is in a similar direction to #85. Could you please review this :).

krassowski:
What is the relation between this (#85) and #86?

Darshan808:
> What is the relation between this (https://github.com/jupyterlab/eslint-plugin/pull/85) and https://github.com/jupyterlab/eslint-plugin/pull/86?

Looking at the docs here, seems like the two PRs are actually aiming for the same thing. so perhaps we should keep just one PR for this rule.


Also, @MUFFANUJ  FYI, I’ve started working on #71. Since some of the rules depend on other rules, I’m raising the PRs one by one to avoid conflicts.

tests/jupyter-require-disposable-ownership.test.ts
Darshan808, line 1: Not sure why there are a lot of code changes in files unrelated to prefer menu helper rule

MUFFANUJ: Ah, I was working/reviewing the possible solution on #89, and it pulled those as well in this one and committed. My bad; it was a branch context mismatch.

MUFFANUJ:
> Looking at the docs here, seems like the two PRs are actually aiming for the same thing. So perhaps we should keep just one PR for this rule.

Yes, but since @Darshan808  had already started working back on #85, I would just close this one! I initially picked it up as it was on draft for over a week, but as this PR would require cleanup, we should probably prioritize the other one.

PR #89, opened 2026-08-04
Fix false positives in disposable ownership rules, improve API

### LINKED ISSUE #87: False positives in the disposable ownership rules
The `require-disposable-ownership` and `require-disposable-transfer` rules added in #77 produce 16 findings on the JupyterLab downstream run. 14 of those are false positives. This issue covers the four causes that are straightforward to fix. Each was confirmed against the JupyterLab source and reproduced as a minimal test case.

| Cause | Downstream findings | Should be fixed before release? |
| --- | --- | --- |
| Exported module singleton reported as unowned | 2 | yes |
| Plugin `activate` written as a named function is not recognised | 5 | yes |
| Disposal inside a callback is not recognised | 2 | yes |
| `ignoredReturnFunctionNames` silently widens the detection scope | 0 (API bug) | yes |
| Ownership through an intermediate options variable | 3 | no, follow-up |
| Ownership handed to a local helper function | 1 | no, follow-up |
| Constructor options map is missing an entry | 1 | no, follow-up |

The two remaining findings are genuine: `packages/toc/src/factory.ts:77` creates an `ActivityMonitor` that is never disposed, and `packages/apputils/src/licenses.tsx:131` creates a `Widget` purely to obtain its `title`.

### 1. Exported module singletons are reported as unowned

Both rules report a disposable assigned to an exported module level binding.

Downstream findings:

- `packages/apputils/src/dialog.tsx:1127`, `export const tracker = new WidgetTracker<Dialog<any>>({ ... })` inside `export namespace Dialog`
- `packages/apputils/src/notification.ts:398`, `export const manager = new NotificationManager()` inside `export namespace Notification`

Reproduction:

```ts
declare class DisposableDelegate { constructor(fn: () => void); }
export const tracker = new DisposableDelegate(() => undefined);
```

An exported binding is created exactly once and lives for the lifetime of the program, so there is nothing to leak. Its disposal, if any is wanted at all, is the responsibility of the importers, which the rule cannot see from this file. Asking the declaring module to own it is not actionable.

Cause: `getAssignedVariable` resolves the variable and the only lifetime guard afterwards is `isOuterFunctionScopeVariable`, which compares the function scope of the creation site against the function scope of the variable. For a top level declaration both are the module scope, so they match and the variable is treated exactly like a function local.

Restricting the exemption to exported bindings clears both downstream findings and keeps all of those tests passing. A broader "any module scope variable" exemption would break 17 existing test cases. 

### 2. Plugin `activate` written as a named function is not recognised

Both rules exempt disposables created directly inside a plugin `activate` function, on the grounds that plugin services usually live for the application lifetime. The exemption only matches one of the three ways `activate` is written, so it misses the most common one in larger extensions.

Downstream findings, all of them disposables created directly in the body of an activation function:

- `packages/apputils-extension/src/palette.ts:98`, `modalPalette` in `Private.activate`
- `packages/csvviewer-extension/src/index.ts:144` and `:311`, `tracker` in `activateCsv` and `activateTsv`
- `packages/toc-extension/src/index.ts:290`, `toolbarMenu` in `activateTOC`
- `packages/console-extension/src/index.ts:340`, `promptMenu` in `activateConsole`

Reproduction:

```ts
declare class DisposableDelegate { constructor(fn: () => void); }
function activateCsv(): void {
  const tracker = new DisposableDelegate(() => undefined);
  void tracker;
}
const plugin = { id: 'x:plugin', autoStart: true, activate: activateCsv };
export default plugin;
```

Cause: `isInJupyterPluginActivate` requires the enclosing function to be the `value` of a `Property` named `activate` on a plugin shaped object literal. That covers `activate: () => { ... }` written inline. It does not cover a separate function referenced as `activate: activateCsv`, and it does not cover `export function activate()` in a `Private` namespace.

### 3. Disposal inside a callback is not recognised

Both rules report a disposable that is disposed inside a callback rather than in the same statement block where it was created.

Downstream findings:

- `packages/apputils/src/thememanager.ts:523`, `splash` is disposed at `:571` inside `requestAnimationFrame(() => { ... })` and at `:576` inside `.catch(reason => { ... })`
- `packages/apputils/src/toolbar/factory.ts:260`, `items` is captured by closures created in `createToolbarFactory` and used after the function returns

Reproduction:

```ts
declare class DisposableDelegate { constructor(fn: () => void); dispose(): void; }
declare function requestAnimationFrame(cb: () => void): void;
function load(): void {
  const splash = new DisposableDelegate(() => undefined);
  requestAnimationFrame(() => {
    splash.dispose();
  });
}
```

Cause: ownership is only credited through `isManagedUse`, which is `isUnconditionalUse` or `isTruthyGuardedUse`. Both stop walking at the first function boundary, so any use inside a callback is invisible. The file already has a narrow workaround for this in `markDisposedSignalCallbackDisposals`, but it is gated to callbacks passed to `<something>.disposed.connect(...)` specifically.

**This in conflict with test cases that were added on purpose**. They assert that this is an error, because `defer` might never run:

```ts
const disposable = createDisposable();
defer(() => {
  disposable.dispose();
});
```

That is structurally identical to `thememanager.ts:568-572`, which is correct code. There is no way to accept one and reject the other, so this is a policy choice rather than a bug.

We can take the position that disposal inside a callback is disposal, on the grounds that a rule shipped at `warn` in the recommended config should prefer a missed detection over a false positive on correct code. To limit what is given up, the fix could only accept unconditional disposal inside the callback, which preserves two of the four affected cases (the ones where the disposal sits inside an `if` within the callback).

### 4. `ignoredReturnFunctionNames` silently widens the detection scope

In `require-disposable-transfer`, setting the `ignoredReturnFunctionNames` option changes far more than the ignore list. Three separate behaviours are gated on whether the option was supplied:

- the exemption for `this._map.set(...)` and `this._map.delete(...)`
- the exemption for names matching `add*Factory`
- the restriction that only factory named calls (`create*`, `build*`, `make*`, `new*`) have their return value checked at all

The first two are defensible, since they are part of the default exemptions and replacing the default list can reasonably replace them too. The third is a detection scope change, not an ignore list change. A user who adds a single name to the ignore list to silence one call site instead gets reports on every call in their codebase whose return type happens to be disposable. Nothing in the rule documentation mentions this.

krassowski:
Closes https://github.com/jupyterlab/eslint-plugin/issues/87

Fixes source of three false positives in `require-disposable-ownership` and `require-disposable-transfer`:
- exported module singletons such as `Dialog.tracker` are owned by the importers of the module, not by the declaring file
- plugin `activate` is now recognised in all three forms it is written in, including a separate function referenced as `activate: activateCsv`
- unconditional disposal inside a callback counts as disposal, so the `requestAnimationFrame(() => splash.dispose())` and promise chain idioms are accepted

This deliberately relaxes one previously asserted behaviour: disposal inside a callback that might never run is no longer reported, which flipped two test cases, while disposal that is itself conditional inside the callback still is.

Improves API:
- both name list options now extend the built-in defaults rather than replacing them, via per-list `extendDefault...` flags,
- adds a standalone `checkAllDisposableReturns`
- removes the array schema `default:` keys that made `['error', {}]` behave differently from `'error'`

Improves documentation by no-longer providing a copy of `DEFAULT_IGNORED_RETURN_FUNCTION_NAMES` in docs to avoid drift.

### DIFF
--- a/src/rules/require-disposable-ownership.ts
+++ b/src/rules/require-disposable-ownership.ts
@@ -15,13 +15,17 @@ import {
   isDisposableConstructor,
   isDisposableExpressionManaged,
   isDisposableType,
+  isDisposedByNestedFunction,
+  isExportedVariable,
   isInJupyterPluginActivate,
   isOuterFunctionScopeVariable,
   markManagedDisposableUse,
-  PendingDisposableMap
+  PendingDisposableMap,
+  resolveNameListOption
 } from '../utils/disposables';
 
 interface RuleOptions {
+  extendDefaultOwnershipFunctionNames?: boolean;
   ownershipFunctionNames?: string[];
 }
 
@@ -47,9 +51,14 @@ const requireDisposableOwnership = createRule({
           ownershipFunctionNames: {
             type: 'array',
             items: { type: 'string' },
-            default: DEFAULT_OWNERSHIP_FUNCTION_NAMES,
             description:
-              'Function or method names that take ownership of a disposable argument.'
+              'Function or method names that take ownership of a disposable argument. Added to the built-in defaults unless extendDefaultOwnershipFunctionNames is false.'
+          },
+          extendDefaultOwnershipFunctionNames: {
+            type: 'boolean',
+            default: true,
+            description:
+              'Whether ownershipFunctionNames extends the built-in defaults. Set to false to replace them, or to drop them entirely when no list is given.'
           }
         },
         additionalProperties: false
@@ -89,9 +98,10 @@ const requireDisposableOwnership = createRule({
       sourceCode: context.sourceCode,
       checker,
       services,
-      ownershipFunctionNames: new Set(
-        (options as RuleOptions).ownershipFunctionNames ??
-          DEFAULT_OWNERSHIP_FUNCTION_NAMES
+      ownershipFunctionNames: resolveNameListOption(
+        (options as RuleOptions).ownershipFunctionNames,
+        DEFAULT_OWNERSHIP_FUNCTION_NAMES,
+        (options as RuleOptions).extendDefaultOwnershipFunctionNames !== false
       )
     };
 
@@ -126,7 +136,9 @@ const requireDisposableOwnership = createRule({
               node,
               assignedVariable,
               context.sourceCode
-            )
+            ) ||
+            isExportedVariable(assignedVariable) ||
+            isDisposedByNestedFunction(assignedVariable, ownership)
           ) {
             return;
           }
--- a/src/rules/require-disposable-transfer.ts
+++ b/src/rules/require-disposable-transfer.ts
@@ -17,14 +17,20 @@ import {
   isDisposableExpressionManaged,
   isDisposableSetFactoryCall,
   isDisposableType,
+  isDisposedByNestedFunction,
+  isExportedVariable,
   isInJupyterPluginActivate,
   isOuterFunctionScopeVariable,
   markManagedDisposableUse,
   PendingDisposableMap,
+  resolveNameListOption,
   shouldCheckReturnedDisposable
 } from '../utils/disposables';
 
 interface RuleOptions {
+  checkAllDisposableReturns?: boolean;
+  extendDefaultIgnoredReturnFunctionNames?: boolean;
+  extendDefaultOwnershipFunctionNames?: boolean;
   ignoredReturnFunctionNames?: string[];
   ownershipFunctionNames?: string[];
 }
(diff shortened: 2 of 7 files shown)


MUFFANUJ (commented):
Is the nested-callback relaxation intentional for callbacks that are only declared but never invoked, registered, or returned? For example, 
```ts 
const cleanupLater = () => disposable.dispose();
``` 
currently appears to mark `disposable` as managed even if `cleanupLater` is never used. If this relaxation is intentional for async cleanup patterns, I think we should still require the callback to actually escape/run in a cleanup-relevant way, otherwise the rule can miss real leaks.

Should conditional ownership/disposal inside callbacks should remain reportable. For example, 
```ts 
defer(() => { if (condition) disposables.add(disposable); })
```
should not mark `disposable` as always managed, because when `condition` is false the disposable is never transferred or disposed. This should behave like the existing same-scope conditional cases and only be accepted when the cleanup/ownership transfer is unconditional within the callback.

Non-blocking: since the issue frames this exemption as covering exported module singletons, should we narrow it to true singleton-style exports such as `export const tracker = ...`? Mutable exported bindings can be reassigned, so treating every exported variable as “created once for the module lifetime” may hide a leak if an exported `let` is replaced and only the latest value is disposed.

MUFFANUJ (approved):
Thanks Mike, this look good!, one thing to note this still misses one issue from #87 false positive: [`packages/apputils/src/toolbar/factory.ts:260`](https://github.com/jupyterlab/jupyterlab/blob/40c26c2ba247966672bb9caba6061d46ca8bc9d1/packages/apputils/src/toolbar/factory.ts#L260).

`createToolbarFactory()` creates `items` once, then returns a factory closure that captures and uses `items` for the lifetime of that returned factory ([lines 279-351](https://github.com/jupyterlab/jupyterlab/blob/40c26c2ba247966672bb9caba6061d46ca8bc9d1/packages/apputils/src/toolbar/factory.ts#L279-L351)). The current nested-callback logic only credits callbacks that directly dispose or hand off the captured disposable, so this returned-factory ownership pattern is still reported as unowned.

we could add a regression test for this shape and extend the exemption to cover disposables captured by an unconditionally returned/escaping factory closure, maybe in a follow up.

MUFFANUJ:
Also, before we merge lets see if Copilot has anything to say since we introduced new behaviour.

PR #91, opened 2026-08-07
Add `no-dynamic-translation` rule

### LINKED ISSUE #90: Catch template literals and dynamic variables in translation rules
### Problem

The plugin may not catch issues with `ITranslator` such a template literals and dynamic variables. For example:

```ts
trans.__(`Delete ${fileName}`);
```

### Proposed Solution

Harden the rule to cover cases like the one noticed in https://github.com/jupyter/notebook/pull/8013

### Additional context

Maybe https://github.com/jupyter/notebook/pull/8013 can be used as a starting point or example to improve the existing rule.

Darshan808:
### Description
`no-translation-concatenation` only caught `+` concatenation in `trans.__()` and missed the cases from jupyter/notebook#8013.

Adds a new rule, **`no-dynamic-translation`**, for everything else that isn't readable at the call site — interpolated templates, variables, calls, member access, conditionals, spreads:

```ts
trans.__(`Delete ${fileName}`);   // interpolation
trans.__(text);                   // variable
```

The two rules split by form, so nothing is reported twice and nothing falls between them. Added at warn in recommended ruleset.

### DIFF
--- a/src/rules/no-translation-concatenation.ts
+++ b/src/rules/no-translation-concatenation.ts
@@ -77,7 +77,7 @@ const noTranslationConcatenation = createRule({
     },
     messages: {
       noConcatenation:
-        'Do not use string concatenation inside translation wrappers. Use a placeholder instead, e.g. trans.__("Hello %1", name).',
+        'Do not concatenate values into translation strings. Use a placeholder instead, e.g. trans.__("Hello %1", name).',
       noInterpolation:
         'Do not interpolate values into translation strings. Use a placeholder instead, e.g. trans.__("Delete %1", fileName).',
       noDynamicMessage:
@@ -107,7 +107,7 @@ const noTranslationConcatenation = createRule({
 
         for (const index of STATIC_ARGUMENT_INDICES[method]) {
           const argument = node.arguments[index];
-          if (!argument || argument.type === 'SpreadElement') {
+          if (!argument) {
             continue;
           }
           if (!isStaticString(argument)) {
--- a/src/rules/no-translation-concatenation.ts
+++ b/src/rules/no-translation-concatenation.ts
@@ -5,7 +5,6 @@
 
 import { createRule } from '../utils/create-rule';
 import { TSESTree } from '@typescript-eslint/types';
-import { TSESLint } from '@typescript-eslint/utils';
 import {
   BUNDLE_METHODS,
   STATIC_ARGUMENT_INDICES,
@@ -13,11 +12,6 @@ import {
 } from '../utils/translation';
 
 type MessageId = 'noConcatenation' | 'noInterpolation' | 'noDynamicMessage';
-
-interface Problem {
-  node: TSESTree.Node;
-  messageId: MessageId;
-}
 
 /**
  * Skips TypeScript-only and optional-chaining wrapper nodes so the underlying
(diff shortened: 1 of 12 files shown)


tests/jupyter-no-translation-concatenation.test.ts
```diff
@@ -29,7 +29,47 @@ ruleTester.run('no-translation-concatenation', noTranslationConcatenation, {
...
+    { code: `trans.__("Delete" as const)` },
+
+    // Variables resolving to a static string are extractable
+    { code: `const MSG = 'Delete'; trans.__(MSG);` },
+    { code: 'const MSG = `Delete`; trans.__(MSG);' },
+    { code: `const MSG = 'a' + 'b'; trans.__(MSG);` },
+
+    // Variables we cannot resolve with confidence are left alone, so generic
+    // translation helpers do not light up
+    { code: `function t(key: string) { return trans.__(key); }` },
```
Darshan808: In cases like this where we cannot resolve variable, it stays silent. Would it be better to flag these as well ?

krassowski: Yes, I think that this should be flagged. It could be a warning or error. Most of the time this will be a mistake (misunderstanding how translator works), the only time it works would be if someone created a static list or translated strings and then passed them untranslated, but I cannot imagine why one would do so.

src/rules/no-translation-concatenation.ts
```diff
@@ -5,105 +5,187 @@
...
+  // Guards against cycles such as `let a = b; let b = a;`.
+  seen.add(variable);
+
+  const written = getSoleWrittenExpression(variable);
+  if (!written) {
+    return null;
+  }
+
+  const problem = classify(sourceCode, written, seen);
+  return problem ? { node: problem.node, messageId: 'noDynamicMessage' } : null;
```
krassowski: Should it flag the usage, not definition?


```suggestion
  return problem ? { node: identifier, messageId: 'noDynamicMessage' } : null;
```

website/docs/rules/no-translation-concatenation.md
```diff
@@ -1,32 +1,70 @@
...
+trans.__('Delete %1', fileName);
+trans._n('%1 file', '%1 files', n);
+
+widget.node.textContent = trans.__('Kernel %1', Text.titleCase(status));
+
+// Still static, so still fine:
+trans.__('Part 1 of long message.\n' + 'Part 2 of long message.\n');
+
+const MESSAGE = 'Delete';
+trans.__(MESSAGE);
```
krassowski: This is not correct at all. This will not be extracted. I think the code is over complicated because it was written under assumption that the definition needs to be studied; it does not - we can just fail if we see any variable usage in the call.

Darshan808:
Total **16** violations for `jupyter/no-translation-concatenation` in lab's codebase.

Darshan808:
In some flagged cases like 
```ts
trans._p('schema', schema.description)
```
`schema.description` is already in the catalog since schemas are extracted and translated from the JSON. These are currently flagged too. Maybe we should add an ignore comment in JupyterLab rather than handle this specific case here. 

Or seems like [schema is translated by the server](https://github.com/jupyterlab/jupyterlab_server/blob/f64f554291a09f072e479ff52ae2212084aaac39/jupyterlab_server/settings_handler.py#L53-L67) before sending it to frontend. So we could even remove that or keep it as a fallback.

krassowski:
> Maybe we should add an ignore comment in JupyterLab rather than handle this specific case here.

Agreed. But let's also document this limitation in the rule description.

> Template handling was merged into this rule instead of creating a new one.

While merging the templates into concatenation is warranted, could we easily extract rejection of non-strings into a separate rule?

Darshan808:
> While merging the templates into concatenation is warranted, could we easily extract rejection of non-strings into a separate rule?

Yes, it can be done easily. So do we want to keep two seperate rule: `no-translation-concatenation` and maybe `no-dynamic-translation` or sth similar?

krassowski:
> do we want to keep two seperate rule: no-translation-concatenation and maybe no-dynamic-translation 

Yes, that sounds good

tests/jupyter-no-translation-concatenation.test.ts
```diff
@@ -19,20 +19,36 @@ const ruleTester = new RuleTester({
 ruleTester.run('no-translation-concatenation', noTranslationConcatenation, {
   valid: [
     { code: `trans.__("Delete %1", fileName)` },
-    { code: `this.trans.__("Hello")` },
-    { code: `this._trans.__("Hello %1", x)` },
-    { code: `this.props.trans.__("Hello")` },
-    { code: `props.trans.__("Hello %1", x)` },
-    { code: `trans.__('Total %1', a + b)` },
```
krassowski, line 26: This could be kept. The others above could be kept too, but not a strong opinion.

krassowski (approved):
Thank you @Darshan808, LGTM!

PR #101, opened 2026-08-14
Add sidebar activity helper rule

MUFFANUJ:
Adds `galata-prefer-sidebar-activity-helper` to encourage Galata tests to use `page.sidebar.openTab()` and `page.activity.activateTab()` instead of raw Playwright selectors for sidebar and main-area tab activation. The change wires the rule into recommended configs, downstream lint config, docs, sidebar navigation, and focused tests, while reusing shared Playwright selector helpers.


src/utils/playwright-selectors.ts
```diff
@@ -42,6 +42,7 @@ export const INTERACTION_METHODS: ReadonlySet<string> = new Set([
  * e.g. `page.locator(sel).first().click()`.
  */
 const CHAIN_PASSTHROUGH_METHODS: ReadonlySet<string> = new Set([
+  'filter',
```
krassowski: I think this needs adjusting to only apply for the sidebar rule. The filebrowser rule would be expanded by this addition with side-effects and unfortunately the test suite did not catch it:
- `page.locator('.jp-DirListing-item').filter({ hasText: 'notebooks' }).dblclick()` would emit `prefer page.filebrowser.open(path)` on a directory which does not work (though maybe we should fix it - feel free to open an issue in lab if you think so) - d86fa62 added `preferOpenDirectory` to avoid it
- `filter({ has: page.locator('.jp-Dialog') })` would report but we added `DIALOG_CONTEXT_PATTERN` to avoid it similar with `filter({ hasNotText: … })`

src/utils/playwright-selectors.ts
```diff
@@ -85,10 +86,36 @@ export interface SelectorInteractionMatch {
...
+    return node.name === 'page' || node.name.endsWith('Page');
+  }
+
+  return (
+    node.type === 'MemberExpression' &&
+    !node.computed &&
+    node.property.type === 'Identifier' &&
+    node.property.name === 'page'
+  );
+}
```
krassowski: Can we use types for this?

website/docs/rules/index.md
```diff
@@ -44,12 +45,14 @@ The plugin ships with a recommended configuration that enables all current rules
 | [jupyter/require-signal-this-arg](./require-signal-this-arg)                                   | `error`  |
 | [jupyter/prefer-signal-this-arg](./prefer-signal-this-arg)                                     | `warn`   |
 | [jupyter/galata-prefer-filebrowser-helper](./galata-prefer-filebrowser-helper)                 | `warn` ¹ |
+| [jupyter/prefer-sidebar-activity-helper](./prefer-sidebar-activity-helper)                     | `warn` ³ |
 | [jupyter/no-schema-enum](./no-schema-enum)                                                     | `warn` ² |
```
krassowski: I would swap 2 and 3 if order dictates it.

src/rules/prefer-sidebar-activity-helper.ts
```diff
@@ -0,0 +1,203 @@
...
+function isGalataHelperImplementation(filename: string): boolean {
+  return filename.replace(/\\/g, '/').includes('/galata/src/helpers/');
+}
+
+function isCloseSidebarComment(value: string): boolean {
+  return /\b(close|closing|collapse|collapsing|hide|hiding)\b/i.test(value);
+}
+
+const preferSidebarActivityHelper = createRule<Options, MessageIds>({
+  name: 'prefer-sidebar-activity-helper',
```
krassowski: It probably should use `galata-` prefix as the existing filebrowser rule does

src/rules/prefer-sidebar-activity-helper.ts
```diff
@@ -0,0 +1,203 @@
...
+  value: string;
+}
+
+const SIDEBAR_TITLE_TO_ID = new Map<string, string>([
+  ['Debugger', 'jp-debugger-sidebar'],
+  ['Extension Manager', 'extensionmanager.main-view'],
+  ['File Browser', 'filebrowser'],
+  ['Property Inspector', 'jp-property-inspector'],
+  ['Running Terminals and Kernels', 'jp-running-sessions'],
+  ['Sessions and Tabs', 'jp-running-sessions'],
```
Darshan808: I don't see `Sessions and Tabs` as title of `jp-running-sessions`. It only contains `Running Terminals and Kernels`. Am I missing something?

src/rules/prefer-sidebar-activity-helper.ts
```diff
@@ -0,0 +1,203 @@
...
+  ['Table of Contents', 'table-of-contents']
+]);
+
+const ACTION_METHODS = new Set(['click', 'tap']);
```
Darshan808: We don't use `page.tap()` in Galata tests. There isn't a single occurrence of it across the test suite, and I don't expect we'll need it. Maybe we could remove it?

src/rules/prefer-sidebar-activity-helper.ts
Darshan808: Other galata helpers are prefixed with `galata-`.
```suggestion
  name: 'galata-prefer-sidebar-activity-helper',
```

src/rules/prefer-sidebar-activity-helper.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    function reportSelectorSource(
+      source: SelectorSource,
+      actionNode: TSESTree.CallExpression
+    ): void {
+      const sidebar = findSidebarTitle(source);
+      if (sidebar) {
+        if (hasNearbyCloseSidebarComment(actionNode)) {
+          return;
+        }
```
Darshan808: I'm not sure about this heuristic of avoiding flags based on a comment containing some specific keywords. I think it would be better to report all sidebar clicks and adjust the message to:

> Use `page.sidebar.openTab("{{ id }}")` to open the `"{{ title }}"` sidebar tab, or `page.sidebar.close(...)` if this click is closing it.

website/docs/rules/prefer-sidebar-activity-helper.md
```diff
@@ -0,0 +1,36 @@
...
+Raw title and text selectors depend on user-facing labels and do not wait for JupyterLab's tab activation state. Galata's helpers target stable tab identifiers or activity names and include the activation checks that make tests less flaky.
+
+## Rule details
+
+The rule reports activation-style calls such as `page.click(selector)` and `page.locator(selector).click()` when the selector directly targets:
+
+- Known sidebar tabs by `title`, such as `Running Terminals and Kernels`, `Property Inspector`, `Table of Contents`, `Extension Manager`, `File Browser`, `Sessions and Tabs`, or `Debugger`.
+- Main area tabs through `div[role="main"] >> text=<tab name>` selectors.
+
+The rule intentionally ignores locator construction and visibility assertions, because those do not necessarily activate a tab. It is intended for Galata test files and skips Galata helper implementation files under `galata/src/helpers/`.
```
Darshan808: `galata/src/helpers/` thing is only for lab. Let's not mention it in general documentation. Maybe remove this whole paragraph.

Darshan808 (changes requested):
Thanks for working on this!

src/rules/prefer-sidebar-activity-helper.ts
MUFFANUJ: You are right; it seems like the LLM added some extra unnecesary once thanks for flagging!

src/rules/prefer-sidebar-activity-helper.ts
MUFFANUJ: Initially, I thought it was a good idea while working on this, but it is kinda very small set where it might be useful; let's remove it, seems like it's overfitting.

src/utils/playwright-selectors.ts
MUFFANUJ: Talking of this, I verified downstream usages across JupyterLab, JupyterLite, and Notebook UI tests. There are no real sidebar/activity helper cases that require following `.filter()` chains, so it feels like removing the `filter`. Please let me know if you think we should add this. For now, I would remove it too.

MUFFANUJ: > * feel free to open an issue in lab if you think so

Sure

krassowski:
@MUFFANUJ can you resolve conflicts and make sure that the new rule is added next to other `galata-` rules in alphabetical order?

krassowski:
As in other PRs locators in `const` are not followed.

```ts
await page.locator('[title="Debugger"]').click();                 // reported
const tab = page.locator('[title="Debugger"]');
await tab.click(); // silent
```



Another issue is miss on `getByRole`:

```ts
await page.click('[title="File Browser"]');                 // reported
await page.getByRole('tab', { name: 'File Browser' }).click(); // silent
```

I will push a few commits for this and other issues.

src/rules/galata-prefer-sidebar-activity-helper.ts
```diff
@@ -0,0 +1,173 @@
...
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'Prefer Galata sidebar and activity helpers over raw Playwright tab selectors',
+      url: 'https://eslint-plugin.readthedocs.io/en/latest/rules/galata-prefer-sidebar-activity-helper/'
+    },
+    messages: {
+      preferSidebarHelper:
+        'Use page.sidebar.openTab("{{ id }}") to open the "{{ title }}" sidebar tab, or page.sidebar.close(...) if this click is closing it.',
```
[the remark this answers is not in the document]
krassowski: It does tak an argument (left or right).

src/rules/galata-prefer-sidebar-activity-helper.ts
```diff
@@ -0,0 +1,445 @@
...
+          : prop.key.type === 'Literal'
+            ? prop.key.value
+            : null;
+      if (typeof key !== 'string' || !NEUTRAL_OPTION_KEYS.has(key)) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
krassowski: That's fine.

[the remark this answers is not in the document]
krassowski: because the risk of that occurring is negligible.

krassowski (approved):
Thanks!

PR #102, opened 2026-08-14
Catch untranslated `label` props via a configurable name list in `no-untranslated-string`

### LINKED ISSUE #100: New heuristic for `no-untranslated-string`: catch all `label` props?
### Problem

A few untranslated strings in JupyterLab were missed, two noted in https://github.com/jupyterlab/jupyterlab/pull/19318:

```
        return new NotebookMetadataField({
          editorFactory,
          tracker,
          label: 'Notebook metadata',
          translator: translator
        }).render(props);
```

### Proposed Solution

Should we flag all `label: string-literal` as possible missing translation?

Maybe this could be an option in `no-untranslated-string`.

### Additional context

None

Darshan808:
### Description

Untranslated `label` props were slipping through since the rule only looked at `addCommand()` and dialog button labels. Instead of a one-off `label` heuristic, the monitored names are now one configurable list, shared by every generic position the rule checks.

```js
{
  "checkProperties": [
    "alt", "aria-description", "aria-label", "caption", "category",
    "label", "placeholder", "title", "textContent", "innerText"
  ]
}
```

One list covers object literal properties, `setAttribute()` attributes, assignment targets and JSX attributes — so `label:` in any object literal, raw `label=` on any element, and `widget.label = 'Save'` / `this.label = 'Save'` are all caught. Also hyphenated and camelCase spellings are the same entry, so `aria-label` and `ariaLabel` are one name.

Also `enforcePunctuation` no longer gates digits, only punctuation, `label: '1970'` is flagged either way.

### DIFF
--- a/src/rules/no-untranslated-string.ts
+++ b/src/rules/no-untranslated-string.ts
@@ -100,24 +100,70 @@ const noUntranslatedString = createRule({
       untranslatedDialogButtonLabel:
         'Dialog button "label" option has an untranslated string literal. Wrap it with trans.__().',
       untranslatedJsxText:
-        'JSX text content has an untranslated string literal. Wrap it with {trans.__(...)}'
+        'JSX text content has an untranslated string literal. Wrap it with {trans.__(...)}',
+      untranslatedLabelProp:
+        '"label" has an untranslated string literal. Wrap it with trans.__().'
     },
     schema: [
       {
         type: 'object',
         properties: {
-          enforcePunctuation: { type: 'boolean' }
+          enforcePunctuation: { type: 'boolean' },
+          checkAllLabels: { type: 'boolean' }
         },
         additionalProperties: false
       }
     ]
   },
-  defaultOptions: [{ enforcePunctuation: false }],
+  defaultOptions: [{ enforcePunctuation: false, checkAllLabels: false }],
 
   create(context) {
-    const enforcePunctuation =
-      (context.options[0] as { enforcePunctuation?: boolean })
-        ?.enforcePunctuation ?? false;
+    const options = context.options[0] as
+      | { enforcePunctuation?: boolean; checkAllLabels?: boolean }
+      | undefined;
+    const enforcePunctuation = options?.enforcePunctuation ?? false;
+    const checkAllLabels = options?.checkAllLabels ?? false;
+
+    // Nodes already reported by a more specific branch (e.g. addCommand or a
+    // Dialog button builder), so the generic `label` check does not duplicate
+    // them. Enclosing calls are visited before the properties they contain.
+    const reportedNodes = new Set<TSESTree.Node>();
+
+    function report(
+      node: TSESTree.Node,
+      messageId:
+        | 'untranslatedCommandProp'
+        | 'untranslatedSetAttribute'
+        | 'untranslatedPropertyAssign'
+        | 'untranslatedTitleProp'
+        | 'untranslatedDialogOption'
+        | 'untranslatedDialogButtonLabel'
+        | 'untranslatedJsxText'
+        | 'untranslatedLabelProp',
+      data?: Record<string, string>
+    ): void {
+      reportedNodes.add(node);
+      context.report({ node, messageId, data });
+    }
+
+    /**
+     * Returns the message id to use for a JSX attribute, or null when the
+     * attribute is not monitored.
+     */
+    function getJsxAttrMessageId(
+      attrName: string | null
+    ): 'untranslatedJsxText' | 'untranslatedLabelProp' | null {
+      if (!attrName) {
+        return null;
+      }
+      if (MONITORED_JSX_ATTRS.includes(attrName)) {
+        return 'untranslatedJsxText';
+      }
+      if (checkAllLabels && attrName === 'label') {
+        return 'untranslatedLabelProp';
+      }
+      return null;
+    }
 
     return {
       CallExpression(node) {
@@ -134,11 +180,7 @@ const noUntranslatedString = createRule({
           for (const propName of MONITORED_COMMAND_PROPS) {
             const prop = properties.get(propName);
             if (prop && isRawStringNode(prop.value)) {
-              context.report({
-                node: prop.value,
-                messageId: 'untranslatedCommandProp',
-                data: { prop: propName }
-              });
+              report(prop.value, 'untranslatedCommandProp', { prop: propName });
             }
           }
           return;
@@ -159,10 +201,8 @@ const noUntranslatedString = createRule({
           }
           const attrValueArg = node.arguments[1];
           if (isRawStringNode(attrValueArg)) {
-            context.report({
-              node: attrValueArg,
-              messageId: 'untranslatedSetAttribute',
-              data: { attr: attrNameArg.value }
+            report(attrValueArg, 'untranslatedSetAttribute', {
+              attr: attrNameArg.value
             });
           }
           return;
@@ -181,10 +221,8 @@ const noUntranslatedString = createRule({
           for (const propName of MONITORED_DIALOG_PROPS) {
             const prop = properties.get(propName);
             if (prop && isRawStringNode(prop.value)) {
-              context.report({
-                node: prop.value,
-                messageId: 'untranslatedDialogOption',
-                data: { prop: propName }
+              report(prop.value, 'untranslatedDialogOption', {
+                prop: propName
               });
             }
           }
@@ -203,10 +241,7 @@ const noUntranslatedString = createRule({
           const properties = getObjectProperties(optionsArg);
           const labelProp = properties.get('label');
           if (labelProp && isRawStringNode(labelProp.value)) {
-            context.report({
-              node: labelProp.value,
-              messageId: 'untranslatedDialogButtonLabel'
-            });
+            report(labelProp.value, 'untranslatedDialogButtonLabel');
           }
         }
       },
@@ -227,11 +262,7 @@ const noUntranslatedString = createRule({
         for (const propName of MONITORED_DIALOG_PROPS) {
           const prop = properties.get(propName);
           if (prop && isRawStringNode(prop.value)) {
-            context.report({
-              node: prop.value,
-              messageId: 'untranslatedDialogOption',
-              data: { prop: propName }
-            });
+            report(prop.value, 'untranslatedDialogOption', { prop: propName });
           }
         }
       },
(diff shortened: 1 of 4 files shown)


src/rules/no-untranslated-string.ts
```diff
@@ -100,24 +100,70 @@ const noUntranslatedString = createRule({
       untranslatedDialogButtonLabel:
         'Dialog button "label" option has an untranslated string literal. Wrap it with trans.__().',
       untranslatedJsxText:
-        'JSX text content has an untranslated string literal. Wrap it with {trans.__(...)}'
+        'JSX text content has an untranslated string literal. Wrap it with {trans.__(...)}',
+      untranslatedLabelProp:
+        '"label" has an untranslated string literal. Wrap it with trans.__().'
```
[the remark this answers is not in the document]
Darshan808: I think it can be left generic.

krassowski (commented):
Shall we enable it for JupyterLab and Notebook downstream testing?

website/docs/rules/no-untranslated-string.md
```diff
@@ -92,12 +92,33 @@ const el = <span>{'raw string'}</span>;
 const el = <span>{trans.__('Error message:')}</span>;
 ```
 
+### 8. Any `label` property (opt-in)
+
+With the [`checkAllLabels`](#checkalllabels) option enabled, every `label: 'raw string'` object property and every raw `label` JSX attribute is flagged, no matter which call or component it belongs to.
```
krassowski: Some high-level thoughts:

1. Should property assignment be flagged too

```ts
widget.label = 'Save'
this.label = 'Save'
```

2. Should it be extended to customizable list of property/attribute names? Maybe a new default could be:

```js
{
  "checkProperties": ["label", "category"],
  "checkJsxAttributes": ["label"],
  "checkAssignments": ["alt", "textContent"]
}
```

Darshan808: I think this makes sense. Let's do this.

Darshan808:
7 new errors got flagged after this enhancement

krassowski:
Real but not actionable:

- packages/filebrowser/src/crumbs.ts
Error:   1164:24  error  Assignment to "textContent" has an untranslated string literal. Wrap it with trans.__()  jupyter/no-untranslated-string
  
  ```ts
  const item = document.createElement('span');
  item.className = BREADCRUMB_SEPARATOR_CLASS;
  item.textContent = '/';
  ```
- packages/debugger/src/sidebar.ts
Error:   234:25  error  Assignment to "textContent" has an untranslated string literal. Wrap it with trans.__()  jupyter/no-untranslated-string
  ```ts
  const title = document.createElement('h2');
  
  title.textContent = '-';
  title.classList.add('jp-text-truncated');
  ```
- packages/rendermime/src/renderers.ts Assignment to "textContent" has an untranslated string literal. Wrap it with trans.__()  jupyter/no-untranslated-string
  ```ts
  anchor.textContent = '¶';
  ```

Real, actionable:

- packages/debugger-extension/src/index.ts
Error:   412:19  error  Property "category" has an untranslated string literal. Wrap it with trans.__()  jupyter/no-untranslated-string
  ```ts
        palette.addItem({
          category: 'Notebook Operations',
          command: Debugger.CommandIDs.restartDebug
        });
  ```
- packages/notebook-extension/src/index.ts
`Error:   1333:14  error  Property "label" has an untranslated string literal. Wrap it with trans.__()  jupyter/no-untranslated-string` and `Error:   1349:14  error  Property "label" has an untranslated string literal. Wrap it with trans.__()  jupyter/no-untranslated-string` (the motivating case)
-  packages/rendermime/src/renderers.ts
Error:    454:7   error  Assignment to "textContent" has an untranslated string literal. Wrap it with trans.__()  jupyter/no-untranslated-string
  ```ts
  if (!trusted) {
    host.textContent =
      'Cannot display an untrusted SVG. Maybe you need to run the cell?';
    return;
  }
  ```
-  packages/rendermime/src/renderers.ts 2092:23  error  Assignment to "alt" has an untranslated string literal. Wrap it with trans.__()          jupyter/no-untranslated-string
  ```ts
  imgs[i].alt = 'Image';
  ```

krassowski:
It looks like non-actionable items are mostly punctuation-like? Should the existing setting for this be respected there?

src/rules/no-untranslated-string.ts
```diff
@@ -93,31 +139,101 @@ const noUntranslatedString = createRule({
...
         additionalProperties: false
       }
     ]
   },
-  defaultOptions: [{ enforcePunctuation: false }],
+  defaultOptions: [
+    {
+      enforcePunctuation: false,
+      checkProperties: DEFAULT_CHECK_PROPERTIES,
+      checkJsxAttributes: DEFAULT_CHECK_JSX_ATTRIBUTES,
```
krassowski: We have a configurable for Jsx attributes and a hard-coded list of attributes. Is this intended?

tests/jupyter-no-untranslated-string.test.ts
```diff
@@ -233,6 +244,241 @@ ruleTester.run('no-untranslated-string', noUntranslatedString, {
...
+        errors: [{ messageId: 'untranslatedProperty', data: { prop: 'label' } }]
+      },
+      // Template literal and concise arrow function
+      {
+        code: `const opts = { label: \`My field\` };`,
+        errors: [{ messageId: 'untranslatedProperty', data: { prop: 'label' } }]
+      },
+      {
+        code: `const opts = { label: () => 'My field' };`,
+        errors: [{ messageId: 'untranslatedProperty', data: { prop: 'label' } }]
```
krassowski, line 313: Low: in arrow functions this should point at the label, not the function.

```suggestion
        errors: [
          {
            messageId: 'untranslatedProperty',
            data: { prop: 'label' },
            type: 'Literal',
            column: 29,
            endColumn: 39
          }
        ]
```

src/rules/no-untranslated-string.ts
krassowski: It looks like the lists diverged too:
- `img.alt = 'A diagram'` reports
- `A diagram` and `el.setAttribute('alt', "A diagram")` do not

krassowski: To have a single list we will need to normalize the hyphenated variant vs camelCase variant (`aria-label` vs `ariaLabel`) but that should be easy.

website/docs/rules/no-untranslated-string.md
```diff
@@ -92,12 +90,73 @@ const el = <span>{'raw string'}</span>;
...
+    "alt",
+    "textContent",
+    "label",
+    "caption"
+  ]
 }
 ```
 
-Set `enforcePunctuation` option to `true` to enforce translation of punctuation characters such as `,`, `-`, `+`, and other symbols.
+Each `check*` option is a list of names that **replaces** the default list rather than adding to it. Pass `[]` to turn that check off entirely.
```
krassowski: Passing `[]` does not turn if all completely as some checks remain hard-coded.

website/docs/rules/no-untranslated-string.md
```diff
@@ -92,12 +90,73 @@ const el = <span>{'raw string'}</span>;
...
+  "checkProperties": ["label", "category"],
+  "checkJsxAttributes": ["aria-label", "aria-description", "title", "label"],
+  "checkAssignments": [
+    "title",
+    "ariaLabel",
+    "alt",
+    "textContent",
+    "label",
+    "caption"
+  ]
```
krassowski: Why not check `alt` and `caption` on JSX? As mentioned earlier maybe we should have just one list here.

The only real reason we would want to keep it separate is that `textContent` has no DOM attribute but maybe that's fine - we would keep API smaller, and make it easier to use?

krassowski: Maybe `textContent` needs to be special? There is also `innerText` that should be handled similarly. If we wanted to be pedantic we could skip checking `textContent` when used on `<style>` or `<script>`.

---

There is also `placeholder` which we should add, JupyterGIS has `placeholder="Enter WMS URL"` untranslated.

src/rules/no-untranslated-string.ts
```diff
@@ -93,31 +139,101 @@ const noUntranslatedString = createRule({
...
+        value.trim().length > 0 && (enforcePunctuation || hasLetters(value))
+      );
+    }
+
+    /**
+     * Returns true if the node is a raw string literal (or a concise arrow
+     * returning one) whose text should be wrapped in a translation call.
+     */
+    function isReportableString(node: TSESTree.Node): boolean {
+      const value = getRawStringValue(node);
```
krassowski: The new checks are bypassed by TypeScript value wrappers

```ts
const o = { label: 'Save' as const };
widget.label = 'Save' as string;
``` 

Worth using `unwrapExpression` here?

https://github.com/jupyterlab/eslint-plugin/blob/8f713684638fa2c2049db1aa7dc6af6ebd731a1b/src/utils/translation.ts#L60-L75

krassowski (commented):
PR title needs updating.

It's a bit awkward that `enforcePunctuation: false` also disables numbers. Maybe that's fine, maybe we can adjust that later.

Darshan808:
Flagged cases mostly looks good.
I think errors like this can be ignored in the lab itself.
```
placeholder: 'https://example.com/path/to/file',
```

src/rules/no-untranslated-string.ts
```diff
@@ -70,12 +92,59 @@ function isDialogButtonCall(node: TSESTree.CallExpression): boolean {
...
+const DEFAULT_CHECK_PROPERTIES = [
+  'alt',
+  'aria-description',
+  'aria-label',
+  'caption',
+  'category',
+  'label',
+  'placeholder',
+  'title',
+  'textContent',
```
krassowski, line 110: One more to add to the default list based on corpus analysis - it is quite common across the ecosystem:

```suggestion
  'textContent',
  'tooltip',
```

website/docs/rules/no-untranslated-string.md
```diff
@@ -92,12 +90,69 @@ const el = <span>{'raw string'}</span>;
...
+  "checkProperties": [
+    "alt",
+    "aria-description",
+    "aria-label",
+    "caption",
+    "category",
+    "label",
+    "placeholder",
+    "title",
+    "textContent",
```
krassowski, line 136:
```suggestion
    "textContent",
    "tooltip",
```

krassowski (commented):
> Also `enforcePunctuation` no longer gates digits, only punctuation, label: '1970' is flagged either way.

It looks like we should not flag bare numbers (as before). I did mention that this is awkward but I was not suggesting that we should change it without further discussion/research.

I imagine the most flexible approach would be adding `enforceNumbers` with default `false`. Or second best keeping it as it was but with a mention in the documentation.

krassowski (approved):
Thank you - approving as this is getting large.

One more edge case to handle in a follow-up PR is conditionals:

```ts
<button title="Show layer" />                              // reported
<button title={visible ? 'Hide layer' : 'Show layer'} />  // not reported
```

PR #103, opened 2026-08-14
Add kernel session types to `longLivedTypes`

### LINKED ISSUE #99: `ISessionContext` and `Session.ISessionConnection` should be in `longLivedTypes`
The `longLivedTypes` list shared by `require-signal-cleanup` and `prefer-signal-this-arg` is currently `CommandRegistry`, `IDebugger`, `IDocumentManager`, `ILSPConnection`, `ILabShell`, `ILanguageServerManager`, `IRenderMimeRegistry`, `ISettingRegistry`, `IShell`, `IStateDB`, `IThemeManager`, `ServiceManager`. Kernel session types are missing, so neither rule can report a connection made to a session signal, whatever the rest of the evidence looks like.

## Why they belong there

A kernel session outlives the widgets built on it. A notebook panel is disposed when its tab is closed, or when a second view of the same document is closed, while the `SessionContext` and its `Session.ISessionConnection` stay alive for as long as the document is open. That is the same argument as for the services already on the list, and it comes up often: JupyterLab core has around 35 `connect()` calls on session or session context signals across 15 files.

## Evidence

Measuring whether closed notebook views are released (5 clone-and-close cycles of a 4 cell notebook, counting live instances with `Runtime.queryObjects` after a forced GC) showed every closed view still reachable. Two of the retaining paths, read off a heap snapshot, run through session signal connections:

1. `ExecutionIndicator.Model` connects to `context.statusChanged` and `context.connectionStatusChanged` with `this` as receiver and is never disposed, because `VDomRenderer.dispose()` drops its reference to the model without disposing it. The connections survive on the `SessionContext`, and the model's `_currentNotebook` field holds the closed view.
2. `DebuggerHandler` connects to `connection.kernelChanged`, `statusChanged`, `iopubMessage` and `anyMessage` with no `thisArg`. Each slot closes over `widget`, so the panel is retained by the session connection.

Both rules are silent on all of these today.

## Scope

Adding the two names removes the first gate, it does not by itself make either rule report the two cases above: `require-signal-cleanup` skips any class with an `extends` clause, and `prefer-signal-this-arg` also requires the class to already clean up by receiver and requires no matching one argument `disconnect` in the file. Those sites may well still be missed, and that is fine. The ask is only that connections to a kernel session become eligible for the remaining gates to judge.

## Proposal

Add `ISessionContext` and `Session.ISessionConnection` to the default `longLivedTypes` in both rules, and re-run the downstream tests to see what the other gates let through.

> Noticed when reviewing https://github.com/jupyterlab/jupyterlab/pull/19307, issue writeup assisted by Opus.

Darshan808:
### Description

Add `ISessionContext` and `ISessionConnection` to the default `longLivedTypes`.

Also make allowlist matching namespace-aware. Entries were compared against the bare symbol name, so a qualified entry like `Session.ISessionConnection` could never match. Entries are now matched as a suffix of the type's declared namespace path: `DocumentRegistry.IContext` matches only that `IContext`, while a bare name still matches any namespace.

### DIFF
--- a/src/utils/signal-lifetime.ts
+++ b/src/utils/signal-lifetime.ts
@@ -53,6 +53,8 @@ export const DEFAULT_LONG_LIVED_TYPES: readonly string[] = [
   'ILabShell',
   'ILanguageServerManager',
   'IRenderMimeRegistry',
+  'ISessionConnection',
+  'ISessionContext',
   'ISettingRegistry',
   'IShell',
   'IStateDB',
@@ -451,11 +453,11 @@ export interface LifetimeContext {
   longLivedTypes: ReadonlySet<string>;
 }
 
-/** Resolves the symbol name of an expression's static type, if available. */
-function typeNameOf(
+/** Resolves the symbol of an expression's static type, if available. */
+function typeSymbolOf(
   node: TSESTree.Node,
   context: LifetimeContext
-): string | null {
+): ts.Symbol | null {
   const { checker, services } = context;
   if (!checker || !services || !services.esTreeNodeToTSNodeMap) {
     return null;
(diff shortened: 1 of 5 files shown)


Darshan808:
Seems like no new classes are being flagged after this enhancement

krassowski (approved):
Thank you!

PR #106, opened 2026-08-26
Add type autofix for schema `oneOf` choices

### LINKED ISSUE #69: `oneOf` requires `type` to render in `rsjf`
## Description

See https://github.com/jupyterlab/jupyterlab/pull/19092. I am not sure if this is the responsibility of this plugin to highlight this but maybe we could add an autofix for `oneOf` which does populate `type: string` when replacing `enum`s?

MUFFANUJ:
Closes #69 

This updates `jupyter/no-schema-enum` to also detect string-valued `oneOf` settings schemas that are missing an explicit `"type": "string"`, which can prevent React JSON Schema Form from rendering the setting control. The rule now reports that case and autofixes it by inserting the missing type before `oneOf`, with tests covering the regression, autofix output, and non-string/typed cases.

### DIFF
--- a/src/rules/no-schema-enum.ts
+++ b/src/rules/no-schema-enum.ts
@@ -10,14 +10,91 @@ type JSONKey =
   | { type: 'JSONLiteral'; value: unknown }
   | { type: 'JSONIdentifier'; name: string };
 
-interface JSONPropertyNode {
-  key: JSONKey;
-  value: { type: string };
+interface JSONNode {
+  type: string;
+  range: [number, number];
   loc: TSESTree.SourceLocation;
+  value?: unknown;
+}
+
+interface JSONPropertyNode {
+  type: 'JSONProperty';
+  key: JSONKey;
+  value: JSONNode;
+  parent?: JSONNode;
+  range: [number, number];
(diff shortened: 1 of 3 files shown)


src/rules/no-schema-enum.ts
```diff
@@ -10,28 +10,108 @@ type JSONKey =
...
+    })
+  );
+}
+
+function getStringTypeInsertion(
+  sourceCode: { text: string },
+  node: JSONPropertyNode
+): string {
+  const beforeNode = sourceCode.text.slice(0, node.range[0]);
+  const lastNewline = beforeNode.lastIndexOf('\n');
```
MUFFANUJ, line 83: Currently, the autofix inserts the `"type": "string"` line with locally computed indentation before "oneOf". Is there a better way to handle formatting here, such as integrating with an existing JSON formatter or relying on another ESLint formatting rule after the fix is applied?

krassowski: I think that's fine. I will push a commit to cover a case when it all stays in a single line (albeit quite unlikely, but for consistency of the output).

krassowski (approved):
Thank you!

PR #108, opened 2026-08-27
Add new `jupyter/prefer-lazy-imports` rule

### LINKED ISSUE #107: New rule: imports of heavy dependencies should happen inside plugin's `activate()` method using `async import`
### Problem

Loading JupyterLab can be slow if large libraries are imported upfront. Oftentimes they are not needed until a user action is performed or at a minimum until the `activate()` method gets called.

### Proposed Solution

Add a rule to encourage moving heavy imports out of the body of a module into a function, a class method, etc.

One heuristic would be to forbid imports from extension code (`./`) or packages that are not expected to be in the shared runtime (i.e. anything that is not `@jupyterlab/` or `@lumino/`, with exception of `@lumino/datagrid`) at the top of `index.ts` which exports the extension plugins. In real world usage plugins can be defined across multiple files so this needs more nuance.

### Additional context

This pattern is used through the codebase already, see https://github.com/jupyterlab/jupyterlab/issues/16135 for examples. It will likely lead to significant wins on cold load time once adopted by third-party extensions.

Also, more related discussion in https://github.com/jupyterlab/jupyterlab/issues/5463.

krassowski:
## References

Closes #107.

## Code changes

- Adds the `prefer-lazy-imports` rule, which reports an import at the top of a plugin module whose bindings are only used inside functions, and points at `await import()` in the function that needs it.
- Adds supporting utilities for the allowlist and the deferred-position walk, for module size measurement, and for reading `jupyterlab.sharedPackages`.
- Adds `typeMentionsJupyterPlugin` and `looksLikePluginObject` to `plugin-utils` for plugin module detection.
- Enables the rule at `warn` in `recommended` and `recommended-legacy`, off for spec and test files, and at `error` in `eslint.downstream.config.mjs`, where `reportInteractionCallbacks` is also enabled.
- Adds the options `allowedPackages`, `ignoreImports`, `minimumSize`, `reportInteractionCallbacks` and `reportModuleLevelUsage`.
- `reportInteractionCallbacks` is opt-in and extends the rule to modules which do not define a plugin, reporting an import only when every use sits inside a user-interaction handler: a command `execute` implementation, a listener for an interaction event such as `click`, or a JSX `on*` handler prop. Those positions cannot run before the user acts and are safe to make `async`.

The defaults come from running the rule over JupyterLab core and about two dozen extensions, roughly 1900 files. The observations below are from that run and will drift as those repositories change.

- `minimumSize` defaults to 4096 bytes. Across that corpus it reports 46% of the imports and covers 95% of the code which could be deferred, against 82% and 99% at 1024, where an async chunk costs about what it saves. The rule docs carry the full table and the asset and cross-file caveats.
- Size is measured by compiling the module with TypeScript and summing the emitted code over its relative import closure. Both halves matter: `shortcuts-extension/src/types.ts` is 7 KB of source and 623 bytes compiled, while `shortcuts-extension/src/renderer.tsx` is 409 bytes on its own and 51 KB once its imports are counted.
- Run through terser and gzip, the relative reports in core total roughly 80 KB minified and 23 KB gzipped. Package reports are worth far more, so they are never filtered by size: `@codemirror/commands` is 82 KB, and `@rjsf/validator-ajv8` pulls in `ajv` at 287 KB.
- Reading `jupyterlab.sharedPackages` is what a fixed allowlist cannot do. JupyterCAD declares `@jupytercad/base` with `bundled: false` in `jupytercad_lab` and `bundled: true` in `jupytercad_core`, so the same import is exempt in one package and reported in the other. This silenced 6 of 26 package reports across the corpus.
- `reportInteractionCallbacks` adds 19 reports on top of the 50 default ones, against 1379 when the plugin-module restriction is lifted with no stricter trigger. Where the reported imports are severable the gain is immediate: deferring `@codemirror/commands` and `@codemirror/search` in `fileeditor-extension/src/commands.ts` takes 49 KB minified (16 KB gzipped) off the startup path. In the jupyterlab-git and JupyterGIS command files every reported module is also reachable through the file's other imports (`FileList` imports `commandsAndMenu` back), so there the reports only pay off once the siblings defer too, which is the single-file limitation the docs describe. The same holds for `@replit/codemirror-vim` (162 KB minified, 51 KB gzipped): the option's report in `labCommands.ts` completes the set of three imports which have to defer together with those the default rule already reports in jupyterlab-vim.

## User-facing changes

Extensions that import heavy dependencies at the top of a plugin module load them before JupyterLab starts, even when the code is only needed after a user action. The rule finds those imports and names the deferred form to replace them with. It stays quiet where deferring cannot help: packages in the shared runtime, tokens in `requires`, `optional` and `provides`, type-only imports, assets the bundler turns into a URL, sources the same file re-exports, and modules too small for a separate chunk to pay off.

## Backwards-incompatible changes

Projects using `configs.recommended` or `recommended-legacy` will see new warnings.

## AI usage

- **YES**: Some or all of the content of this PR was generated by AI.
- **YES**: The human author has carefully reviewed this PR and run this code (keep this PR "draft" until the answer is YES)
- AI tools and models used: Opus 5, Fable 5 for review


krassowski:
## Example impact on load time

Share of each project's own compiled code that the rule reports as deferrable:

| project                       | own code | deferrable | share |
| ----------------------------- | -------- | ---------- | ----- |
| jupyterlab-spreadsheet-editor | 152 KB   | 113 KB     | 74%   |
| jupyter-resource-usage        | 120 KB   | 72 KB      | 60%   |
| jupyterlab-variableInspector  | 94 KB    | 53 KB      | 56%   |
| jupyterlab-vim                | 30 KB    | 12 KB      | 40%   |
| jupyterlab_code_formatter     | 35 KB    | 12 KB      | 35%   |
| jupyterlab-git                | 4.4 MB   | 848 KB     | 19%   |
| jupyterlab core               | 13.5 MB  | 103 KB     | 0.8%  |

A single-purpose extension is mostly one widget behind one command, so most of it can wait for a user action. Core is the shared runtime and most of it is needed at startup.

The load time cost of those bytes (each figure is the extra page load time the payload adds over an empty baseline):

| project                       | deferrable | 3G       | Slow 4G | Fast 4G |
| ----------------------------- | ---------- | -------- | ------- | ------- |
| jupyterlab-vim                | 12 KB      | 235 ms   | 57 ms   | 11 ms   |
| jupyterlab_code_formatter     | 12 KB      | 236 ms   | 57 ms   | 11 ms   |
| jupyterlab-variableInspector  | 53 KB      | 1059 ms  | 258 ms  | 47 ms   |
| jupyter-resource-usage        | 72 KB      | 1440 ms  | 366 ms  | 69 ms   |
| jupyterlab core               | 103 KB     | 2059 ms  | 516 ms  | 96 ms   |
| jupyterlab-spreadsheet-editor | 113 KB     | 2261 ms  | 564 ms  | 104 ms  |
| jupyterlab-git                | 848 KB     | 16983 ms | 4169 ms | 767 ms  |

Measured in Chromium through Playwright with the DevTools throttling presets, serving real built extension code with the cache disabled, median of three runs. These above are uncompressed (which is what Jupyter Server sends).

Behind a proxy that adds compresion the same payloads cost roughly a quarter as much:

| project                       | deferrable | 3G      | Slow 4G | Fast 4G |
| ----------------------------- | ---------- | ------- | ------- | ------- |
| jupyterlab-vim                | 12 KB      | 58 ms   | 16 ms   | 6 ms    |
| jupyterlab_code_formatter     | 12 KB      | 57 ms   | 16 ms   | 0 ms    |
| jupyterlab-variableInspector  | 53 KB      | 201 ms  | 41 ms   | 9 ms    |
| jupyter-resource-usage        | 72 KB      | 265 ms  | 56 ms   | 16 ms   |
| jupyterlab core               | 103 KB     | 380 ms  | 85 ms   | 19 ms   |
| jupyterlab-spreadsheet-editor | 113 KB     | 407 ms  | 91 ms   | 17 ms   |
| jupyterlab-git                | 848 KB     | 2963 ms | 717 ms  | 141 ms  |

Darshan808 (approved):
This looks good to me. The flagged cases appear to be true positives, and the documentation is comprehensive enough for anyone who wants to understand how the rule works in depth.
Thanks!

PR #122, opened 2026-09-16
Show better message for heavy imports in autostart plugins, improve docs

### LINKED ISSUE #121: `prefer-lazy-imports`: the suggested fix is wrong for imports used only in an autostart `activate()`
### Problem

The rule reports an import used only inside `activate()` of an `autoStart: true` plugin with the same message as any other:

```
'./kernels' is imported at the top of a plugin module but only used inside functions, so it loads before the application starts. Import it where it is used instead: `const { addKernelRunningSessionManager } = await import('./kernels');`
```

Following that snippet inside `activate()` does not help and can hurt: `Application.start` awaits every autostart plugin before attaching the shell, so the chunk still loads before the application starts, one request later, and any `await` that follows the import (a service `ready`, a settings load) now blocks the first paint. In jupyterlab/jupyterlab#19751 this turned a 0.9 s shell attach into 3.4 s when one request was slow, and across the extensions I linted 92 of 138 reports were of this kind.

### Example

```ts
import { addKernelRunningSessionManager } from './kernels';

const plugin: JupyterFrontEndPlugin<IRunningSessionManagers> = {
  id: 'running-extension:plugin',
  autoStart: true,
  activate: app => {
    const managers = new RunningSessionManagers();
    void addKernelRunningSessionManager(managers, app);
    return managers;
  }
};
```

### Proposed Solution

When every use of the import sits inside `activate()` of a plugin with `autoStart: true`, report a different message:

```
'./kernels' is imported at the top of a plugin module and used only in `activate()` of an autostart plugin, so it loads before the application starts either way. Do not `await import(...)` inside `activate()`: that delays the whole application start. Register the extension point synchronously and load './kernels' in the callback that first needs it, or ignore this import if activation needs it at once.
```

krassowski:
## References

Closes #121.

## Code changes

- Replaces `isEagerlyReached` with `getReach`, which places a use at `module`, `activation` or `deferred`. A use is at `activation` when it runs inside `activate` of a plugin object with `autoStart: true`: written in place, named as `activate: fn` or in the shorthand `{ activate }`, or inside a helper which `activate` calls, followed transitively as before.
- Adds the `usedInAutostartActivate` message and reports it, without the `await import()` snippet, when no use is eager and at least one runs during activation. A use in a command, in an `app.restored` callback, or in `activate` of a plugin without `autoStart` or with `autoStart: 'defer'` keeps the `preferLazyImport` message.
- Adds a test block for the autostart cases and their near misses.
- Rewrites the correct example in the rule docs to defer inside a command instead of making `activate` async, and adds an "Autostart plugins" section which contrasts an `await import()` at the top of `activate()` with the same import inside the `execute` callback.

The set of reported imports does not change. An import used in an autostart `activate` was reported before as well, with advice that made things worse.

## User-facing changes

Before, an import used only inside `activate()` of an `autoStart: true` plugin got the same message as any other, ending in `const { X } = await import('./x');`. Following that inside `activate()` keeps the request on the startup path and makes `Application.start` wait for it, because the shell is attached only after every autostart plugin has activated. In [jupyterlab/jupyterlab#19751](https://github.com/jupyterlab/jupyterlab/pull/19751) this turned a 0.9 s shell attach into 3.4 s when one request was slow, and 92 of the 138 reports across the linted extensions were of this kind.

Now such an import is reported as:

> './kernels' is imported at the top of a plugin module and used in `activate()` of an autostart plugin, so it loads before the application starts either way. Do not `await import(...)` inside `activate()`: that delays the whole application start. Register the extension point synchronously and load './kernels' in the callback that first needs it, or ignore this import if activation needs it at once.

One use in `activate()` is enough for this message, even when the other uses sit in callbacks, because the module is fetched at start either way and the snippet would be wrong for the use in `activate()`.

## Backwards-incompatible changes

None

## AI usage

- **Yes**: Some or all of the content of this PR was generated by AI.
- **Yes**: The human author has carefully reviewed this PR and run this code (keep this PR "draft" until the answer is YES)
- AI tools and models used: Fable 5.1


krassowski:
I'm going to ignore this spree that Copilot took in pointing out various imaginary ways that a plugin can be defined. Yes, this means that there will be some false negatives, but it is better than adding hundreds of lines of code just to cover edge cases that do not appear in the corpus of hundreds of plugins that we already maintain.

MUFFANUJ (approved):
Thanks!

PR #125, opened 2026-09-19
Make rule documentation easier to read

### LINKED ISSUE #117: Make the rule documentation easy to read for humans
### Problem

The rule docs often include implementation details upfront. There is too much text, too many details and too many code symbols.

### Proposed Solution

- Put examples (incorrect/correct) first, humans parse examples better
- Add more examples for motivation
- Push implementation details down to `<details>` section unless relevant to a human user (not maintainer)
- Drop implementation details that are only relevant to maintainers if they are already well described in the tests and code comments, barring high-level design remarks

MUFFANUJ:
Closes #117 

Reorganizes all 23 rule pages to put incorrect/correct examples first, explain user impact clearly, and move secondary technical details into collapsible sections. Removes redundant implementation notes while preserving configuration guidance and practical caveats. Adds documentation guidance for contributors. Documentation-only change; rule behavior is unchanged.


website/docs/rules/galata-prefer-context-menu-helper.md
```diff
@@ -63,6 +29,17 @@ await page.filebrowser.open('README.md', 'Markdown Preview');
 await page.filebrowser.open('data.csv', 'CSV Viewer');
 ```
```
MUFFANUJ: This looks obvious to me but for a user who is reading this it might be difficult to understand what is wrong here and what is correct instead of having one incorrect section showing all incorrect implementation and correct showing correct implementation will it make sense to have multiple incorrect - correct sections to understand better whats wrong and whats right i guess we could do this for most of the rules?

krassowski: Yes, probably better to split and name the combined section "Examples"

website/docs/rules/require-signal-cleanup.md
krassowski:
```suggestion
- `additionalCleanupMethods` (`string[]`, default `[]`): additional method names (besides `disconnect`) that count as cleanup evidence when called anywhere in the class. Use this to allowlist project-specific teardown idioms.
```

website/docs/rules/command-described-by.md
```diff
@@ -2,19 +2,13 @@
...
-
-## Rule details
+## Examples
 
-The rule checks calls that match `*.addCommand(...)` patterns and reports when:
+### Describe a command argument
 
-- The command options object has an `execute` function
-- But does not provide `describedBy`
+The `value` argument needs metadata describing the value callers should provide.
```
krassowski: A bit easier to read

```suggestion
The `value` argument needs metadata that describes the value a caller should pass.
```

website/docs/rules/command-described-by.md
```diff
@@ -44,6 +38,10 @@ app.commands.addCommand(CommandIDs.test, {
 });
 ```
 
+## Why
+
+Command argument metadata tells callers which values a command accepts. For example, the schema above describes `value` as a string. Add `describedBy` when registering a command with an `execute` function.
```
krassowski: This is mostly correct but is not very useful as motivation.

Also, a command without `execute` is not a thing really - maybe technically possible but not worth mentioning as it would be useless. Something like:

```suggestion
The commands are a fundamental building block in Jupyter frontend applications. They define actions reused by menus, shortcuts and command palette. Commands are also reused by extension authors and AI integrations alike.

Adding a description of the arguments helps extension authors develop better integrations and LLMs call the command with the right shape of arguments. Certain LLM extensions map the commands to tool calls 1:1.

The schema-based argument description is also used programmatically in JupyterLab and extensions:
- in Keyboard Shortcuts UI allowing to tweak the action invoked on a shortcut press
- in ui-profiler extension
- in plugin-playground
```

website/docs/rules/galata-prefer-context-menu-helper.md
```diff
@@ -1,68 +1,75 @@
...
 await page.click('.lm-Menu-itemLabel:text("Markdown Preview")');
+```
+
+**Correct**
 
+```ts
+await page.filebrowser.open('README.md', 'Markdown Preview');
+```
+
+### Choose CSV Viewer using role locators
```
krassowski: Well, the correct example has no locators:

```suggestion
### Choose CSV Viewer
```

website/docs/rules/galata-prefer-filebrowser-helper.md
```diff
@@ -2,49 +2,96 @@
...
+
+Raw file browser selectors depend on markup that can change and can skip readiness checks. The helpers open intermediate directories and wait for the document tab; `page.notebook.openByPath()` also waits for the notebook panel to be ready.
+
+Use `openDirectory()` when opening a directory itself. `open()` waits for a document tab, which opening a directory does not create.
+
 ## Options
 
 This rule has no options.
+
+## Configuration
```
krassowski: What is below could be described better as 

```suggestion
## Limitations
```

website/docs/rules/galata-prefer-filebrowser-helper.md
```diff
@@ -2,49 +2,96 @@
...
+
+Use the Galata `page` fixture to make these helpers available. The rule also reports raw interactions on a plain Playwright fixture named `page`; disable it for files where you intentionally use that fixture.
```
krassowski:
```suggestion
To make these helpers available you must use the Galata `page` fixture. The rule also reports raw interactions on a plain Playwright fixture named `page`; you may prefer to disable it for files where you intentionally use the plain Playwright fixture.
```

website/docs/rules/galata-prefer-filebrowser-helper.md
```diff
@@ -2,49 +2,96 @@
...
+
+**Allowed**
+
+```ts
 await page.locator('.jp-Dialog .jp-DirListing-itemName').first().dblclick();
 ```
 
+## Why
+
+Raw file browser selectors depend on markup that can change and can skip readiness checks. The helpers open intermediate directories and wait for the document tab; `page.notebook.openByPath()` also waits for the notebook panel to be ready.
```
krassowski:
```suggestion
Raw file browser selectors depend on markup that can change, and they skip the readiness checks the helpers perform. The helpers open intermediate directories and wait for the document tab; `page.notebook.openByPath()` also waits for the notebook panel to be ready.
```

website/docs/rules/galata-prefer-filebrowser-helper.md
```diff
@@ -2,49 +2,96 @@
...
+
+**Correct**
+
+```ts
+await page.filebrowser.open('data.json');
+```
+
+### File selector dialogs
+
+The file browser helper drives the sidebar, so it cannot replace interactions in a file selector dialog.
```
krassowski:
```suggestion
The file browser operates on the sidebar, so it cannot replace interactions in a file selector dialog.
```

website/docs/rules/galata-prefer-menu-helper.md
```diff
@@ -2,81 +2,80 @@
...
+## Why
+
+Raw clicks depend on which menu is already open and when its submenus appear. `page.menu.clickMenuItem()` closes open menus, follows the requested path and waits for each submenu. This makes a sequence such as opening a terminal easier to read and less prone to timing failures.
 
-```ts
-await page.menu.clickMenuItem('File>New>Terminal');
-await page.menu.openLocator('File');
-await page.menu.closeAll();
-```
+Use `page.menu.isOpen(path)` to check menu state and `page.menu.getMenuItemLocator(path)` to inspect an item. Prefer `openLocator()` and `getMenuItemLocator()` over the deprecated `open()` and `getMenuItem()` forms.
```
krassowski: Does it belong in Why?

website/docs/rules/galata-prefer-notebook-cell-helper.md
```diff
@@ -2,81 +2,98 @@
...
-await page.notebook.enterCellEditingMode(0);
 await page.notebook.selectCells(2);
+```
 
-// Not a notebook cell: the console reuses the same editor markup.
+### Console input
+
+The console reuses editor markup but is not a notebook cell.
+
+**Allowed**
```
krassowski:
```suggestion
**Not reported**
```

website/docs/rules/galata-prefer-notebook-cell-helper.md
```diff
@@ -2,81 +2,98 @@
...
 
+## Why
+
+Raw cell selectors depend on notebook markup and skip the readiness checks built into Galata. The helpers describe the action directly: set the source, select a cell or run it.
+
 ## Options
 
 This rule has no options.
+
+## Choosing a helper
```
krassowski:
```suggestion
## When not to use it
```

website/docs/rules/incorrect-translator-usage.md
```diff
@@ -1,55 +1,89 @@
 # `incorrect-translator-usage`
 
-Require translation bundles returned by `translator.load()` to be stored under an extractor-recognized name (e.g. `trans`, `this.trans`, `this._trans`, `props.trans`, `this.props.trans`).
+Store translation bundles under a name the translation extractor recognizes, such as `trans`.
 
-## Why
+## Examples
 
-JupyterLab collects translatable strings statically with a gettext-based extractor. The extractor only recognizes translation calls made through a small set of names:
+### Keep the bundle before translating
```
krassowski:
```suggestion
### Store the bundle in a variable
```

website/docs/rules/prefer-lazy-imports.md
```diff
@@ -70,99 +43,155 @@ const plugin: JupyterFrontEndPlugin<void> = {
...
+
+export async function createGrid(): Promise<DataGrid> {
+  const { DataGrid } = await import('@lumino/datagrid');
+  return new DataGrid();
+}
+```
+
+### Load an application-specific class during on-demand activation
+
+This plugin is activated on demand and does not declare `autoStart: true`. Importing the Notebook application package only for an instance check otherwise adds it to startup work. Review dependencies from autostart plugins too; see [Autostart plugins](#autostart-plugins).
```
krassowski:
```suggestion
This plugin is activated on demand and does not declare `autoStart: true`. A static import of the Notebook application package, used only for an instance check, loads that package at startup. Review dependencies from autostart plugins too; see [Autostart plugins](#autostart-plugins).
```

website/docs/rules/prefer-lazy-imports.md
```diff
@@ -326,35 +266,15 @@ The default is the singleton list from JupyterLab's `staging/package.json`, minu
...
-}
-```
-
-That example keeps five of the defaults and adds `@myorg/*`. The eleven defaults it leaves out, among them `@jupyter/ydoc`, `yjs` and the three `@codemirror` packages, are reported again.
+Packages declared with `bundled: false` in the extension's `jupyterlab.sharedPackages` manifest are also exempt, unless listed in `deferredPackages`. Changing `allowedPackages` does not remove these manifest exemptions.
 
 ### `deferredPackages`
 
-Type: `string[]`, default: the list below.
+Packages that should load on demand. Static imports and re-exports are reported in every file, including side-effect imports and imports with no runtime use. Whole `import type` and `export type` declarations are exempt, as are the asset types described below.
```
krassowski:
```suggestion
Packages that should load on demand. Static imports and re-exports are reported in every file, including side-effect imports and imports with no runtime use. Whole `import type` and `export type` declarations are exempt, as are the asset types described below in "Build assumptions, assets and shared packages" details section.
```

website/docs/rules/prefer-lazy-imports.md
```diff
@@ -440,85 +332,75 @@ The table below comes from a run over JupyterLab core and about two dozen extens
...
 
-```json
-{
-  "reportModuleLevelUsage": true
-}
-```
+Plugin files include declarations using `JupyterFrontEndPlugin` or `ServiceManagerPlugin`, including arrays, unions, promises, casts and factory return types. Untyped objects count when they contain a string `id`, an `activate` function and at least one of `autoStart`, `requires`, `optional`, `provides` or `description`. Aliased type names need type information.
 
-## Known limitations
+For packages outside `deferredPackages`, the default checks imports used only inside functions, methods or instance field initializers. It skips type-only uses, side-effect imports, plugin tokens and sources re-exported from the same file. Imports needed during module evaluation, including through a helper called at module scope, are left alone unless `reportModuleLevelUsage` is enabled. An eager re-export keeps the source in the startup bundle even if its other uses are deferred.
```
krassowski:
```suggestion
For packages outside `deferredPackages`, by default the rule checks imports used only inside functions, methods or instance field initializers. It skips type-only uses, side-effect imports, plugin tokens and sources re-exported from the same file. Imports needed during module evaluation, including through a helper called at module scope, are left alone unless `reportModuleLevelUsage` is enabled. An eager re-export keeps the source in the startup bundle even if its other uses are deferred.
```

website/docs/rules/prefer-signal-this-arg.md
```diff
@@ -1,156 +1,135 @@
...
-```ts
-// The sender is constructed and disposed here, so it dies with the receiver
-// and the missing thisArg costs nothing.
-class Host extends Widget {
-  constructor() {
-    super();
-    this._editor = createEditor();
-    this._editor.ready.connect(() => this.update());
-  }
+## Why
```
krassowski: It should probably mention the danger of memory leaks and degradation of UX when these accumulate (a stale noot proparly disposed signal can end up holding the entire widget alive)

website/docs/rules/require-signal-this-arg.md
```diff
@@ -2,70 +2,102 @@
 
 Require a `thisArg` when connecting a class method that references `this` to a Lumino signal.
```
krassowski: Require `this` as the second argument when a class method that uses `this` is connected to a Lumino signal.

website/docs/rules/require-soft-assertions-before-snapshots.md
```diff
@@ -2,48 +2,98 @@
...
 ```ts
 test('single screenshot', async ({ page }) => {
-  expect(await page.screenshot()).toMatchSnapshot('page.png'); // only snapshot — fine
+  expect(await page.screenshot()).toMatchSnapshot('page.png');
 });
 ```
 
+## Why
+
+A hard snapshot assertion stops the test at its first failure. Later screenshots are never captured, so you cannot inspect all visual changes from that run. Use `expect.soft()` to continue to the remaining snapshots while still failing the test if any assertion fails. The last snapshot can use either form. This also allows later snapshots to be reached when running with `--update-snapshots`.
```
krassowski:
```suggestion
A hard snapshot assertion stops the test at its first failure. Later screenshots are never captured, so you cannot inspect all visual changes from that run. Use `expect.soft()` to continue to the remaining snapshots while still failing the test if any assertion fails. The last snapshot can use either form. Soft assertions also let `--update-snapshots` reach every snapshot in one run.
```

website/docs/rules/token-format.md
```diff
@@ -29,6 +26,45 @@ export const IFooService = new Token<IFooService>(
...
+
+**Correct**
+
+```ts
+new Token<IVersion2Service>('@test/pkg:IVersion2Service', 'Version 2 service');
+```
+
+## Why
+
+Including the package name and service symbol makes a token easy to identify. The part after `:` must start with an ASCII letter, `_` or `$` and contain only those characters or digits. A missing `:` is also reported.
```
krassowski: Move scope details to scop

```suggestion
Including the package name and service symbol makes a token easy to identify.
```

website/docs/rules/token-format.md
```diff
@@ -29,6 +26,45 @@ export const IFooService = new Token<IFooService>(
...
+
 ## Options
 
 This rule has no options.
+
+<details>
+<summary>Which token IDs are checked?</summary>
+
+The rule checks string literals passed to `new Token(...)`. Variables and template literals are not checked.
```
krassowski:
```suggestion
The rule checks string literals passed to `new Token(...)`. Variables and template literals are not checked.

The part after `:` must start with an ASCII letter, `_` or `$` and contain only those characters or digits. A missing `:` is also reported.
```

krassowski (commented):
Reads much better, some suggestions left

krassowski:
@Darshan808 have you had a chance to take a look? If not we could always change things in af follow up, I just don't want to merge if you are mid review

Darshan808 (approved):
Looks good, thanks!

